PR #3, opened 2015-04-02
[WIP] UX improvements

joostdevries:
Three tabs:
- JS
- HBS
- Output

Still a lot to do but a first idea...

image

demo: http://twiddle.joostdvrs.com/f2c90713bdfdf54a262b
Right now login, post/update don't work on the demo (because we would then need to update the app on github again). To check those, clone locally.

@alexspeller @stefanpenner


joostdevries:
Superseded by #8

PR #23, opened 2015-06-03
Update to ember-cli 0.2.7

inDream:
Fix Object.defineProperty called on non-object


joostdevries:
Thanks for this, I've extracted some of the deps and updated master to use Ember 1.13.

Issue #21, opened 2015-05-31
Build into iframe


joostdevries:
Instead of 'lazily' builing the app file-by-file on change using the resolver, run a full build and dump the code into an iframe.

Benefits:
- isolate the app from the editor
- allows other versions of ember to be used
- use css without polluting parent window

### modules we need to include
- app/app.js
- app/router.js
- boot.js (boots the app)

### EmberCLIService
- buildGist
- fileFromBlueprint
- contentForAppBoot

joostdevries:
closed in #24

PR #24, opened 2015-06-14
Refactor build process to build into iframe

joostdevries:
- [x] Ember 1.13
- [x] Build the app into an iframe #21


ef4:
This is looking nice.

Just so you aren't surprised: blueprints are likely to move from ember-cli to ember pretty soon.

joostdevries:
@ef4 cool. on what term and where to? If they live in an addon that's fine, I'll still be able to import them.. I just hate having duplicate anything.

PR #26, opened 2015-06-23
Test cleanup

piotrpalek:
Hey I cleaned the tests up a bit so they pass again, I wasn't sure how to port the unused `gist-test` so I deleted it for now. The reason I did this is I wanted to try and contribute a bit to this repository but also write some tests at the same time.


piotrpalek:
@alexspeller don't want to be nagging but.. what do you think about this?

joostdevries:
thanks!

PR #27, opened 2015-06-24
Fix problem when removing file opened in 2 columns

### LINKED ISSUE #19: Handle delete of file in secondary column
In the primary column we set file to null but if the same file is opened in both columns, the secondary col. still shows the file which leads to issues.

piotrpalek:
This fixes #19 and is a folloup to #26. It also adds better behavior when removing files. When a file is removed the file selector now shows "No file selected", and if all files are removed the dropdown is hidden and "No files available" is shown.


joostdevries:
thanks!

Issue #11, opened 2015-05-17
Support routing


joostdevries:
- [ ] Support Twiddling with the Router
- [ ] Mock a navbar

Gaurav0:
This is fixed.

PR #51, opened 2015-07-13
Enable routing (fixes issue #11)

Gaurav0:
This adds a file named 'router.js' with a default implementation that can be used as the router to the Ember Twiddle default blueprint app

Fixes issue #11


app/lib/twiddle-resolver.js
```diff
@@ -6,50 +6,44 @@ export default Resolver.extend({
   },
 
   resolveOther (parsedName) {
-    if (parsedName.fullName === 'router:main') {
```
joostdevries: `twiddle-resolver` doesn't do anything anymore as the application is now using the normal resolver

app/gist/new/route.js
```diff
@@ -13,12 +13,16 @@ export default GistRoute.extend({
...
-      content: '<h1>Welcome to {{appName}}</h1><br><br>',
+      content: '<h1>Welcome to {{appName}}</h1><br><br>{{outlet}}<br><br>',
     }));
     model.get('files').pushObject(this.store.createRecord('gistFile', {
       filePath: 'controllers/application.js',
       content: 'import Ember from \'ember\';\n\nexport default Ember.Controller.extend({\n  appName:\'Ember Twiddle\'\n});',
     }));
+    model.get('files').pushObject(this.store.createRecord('gistFile', {
+      filePath: 'router.js',
+      content: 'import Ember from \'ember\';\nvar Router = Ember.Router.extend({\n  location: \'none\'\n});\n\nRouter.map(function() {\n});\n\nexport default Router;\n'
```
joostdevries: I'd rather have an actual router with an address bar (although it will probably have to be readonly)

Gaurav0: I can add an address bar if you'd like, but I'm not sure what you mean by an actual router?

app/lib/twiddle-resolver.js
Gaurav0: I didn't know that. Should I delete it?

app/gist/new/route.js
joostdevries: sorry I mean actual location.

app/lib/twiddle-resolver.js
joostdevries: That's cool.

app/gist/new/route.js
Gaurav0: Address bar added.

stefanpenner:
YAY, good job @Gaurav0

joostdevries:
:heart_eyes:

PR #54, opened 2015-07-15
Support ember data

Gaurav0:


joostdevries:
:+1:

stefanpenner:
ah, was that all that was needed?

joostdevries:
@stefanpenner yeah probably due to the Esperanto/Babel difference

Issue #55, opened 2015-07-15
Could not find module `demo-app/router` imported from `demo-app/initializers/router`


aortbals:
A [twiddle](http://ember-twiddle.com/245f05c2105f92768d04) from a few days ago that I made was working great - love the tool btw - but I'm now getting this error:

> Uncaught Error: Could not find module `demo-app/router` imported from `demo-app/initializers/router`

aortbals:
Must be related to https://github.com/ember-cli/ember-twiddle/pull/51

Gaurav0:
So the problem is there is now a default router.js required. Please create a router.js containing the following:

    import Ember from 'ember';
    var Router = Ember.Router.extend({
      location: 'none'
    });
    
    Router.map(function() {
   });
   
   export default Router;

Gaurav0:
The file is created with new twiddles, but existing twiddles will need to add this. Please leave this open to see if we can come up with something that inserts it automatically into old twiddles.

aortbals:
Indeed that did fix it. I agree though, it would be unfortunate if existing twiddles remaining broken from this change.

joostdevries:
fix by @Gaurav0 has been deployed :-).

aortbals:
Thanks @Gaurav0!

PR #56, opened 2015-07-15
Make router.js optional; add router blueprint (fixes issue #55)

Gaurav0:
This allows existing Ember Twiddle gists without a router.js to continue to work by making router.js optional. Users can now add a router through a blueprint. This also lets them conveniently reset the router by deleting and adding a new one.

Fixes issue #55

### DIFF
--- a/app/services/ember-cli.js
+++ b/app/services/ember-cli.js
@@ -48,6 +48,7 @@ export default Em.Service.extend({
       // Add router
       if (!gist.get('files').findBy('nameWithModule', 'demo-app/router')) {
         out.push(this.compileJs(blueprints.router, 'demo-app/router'));
+        console.error('Missing a router.js See issue https://github.com/ember-cli/ember-twiddle/issues/55#issuecomment-121724317');
       }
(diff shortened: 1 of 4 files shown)


Gaurav0:
Added an error message in the console (not shown in browser).

app/services/ember-cli.js
```diff
@@ -42,9 +42,17 @@ export default Em.Service.extend({
...
       if (errors.length) {return reject(errors);}
 
-      // Add app, config
+      // Add app
       out.push(this.compileJs(blueprints.app, 'demo-app/app'));
+
+      // Add router
+      if (!gist.get('files').findBy('nameWithModule', 'demo-app/router')) {
+        out.push(this.compileJs(blueprints.router, 'demo-app/router'));
+        console.error('Missing a router.js See issue https://github.com/ember-cli/ember-twiddle/issues/55#issuecomment-121724317');
```
stefanpenner: warn or error?

Gaurav0: @joostdevries @alexspeller Care to weigh in?

Gaurav0: Actually, I have a better idea. If we added a blueprint, we could make router.js permanently optional.

joostdevries:
@Gaurav0 thanks! Ultimately I think we can refactor this to be a bit more abstract+concise:
- all blueprints should come from lib/blueprints. 
- some blueprints should be marked as `required`. if they are not present in the filetree, we'll add them at build time.

Issue #30, opened 2015-07-08
Optionally disable auto-run


joostdevries:
We should be careful w/ auto-run. Sometimes people make reproductions of infinite loops etc

stefanpenner:
can someone share an example of this? I don't think we want to actually turn off auto-runs..

stefanpenner:
> can someone share an example of this? I don't think we want to actually turn off auto-runs..

you know, i totally assumed this was related to embers run-loop autoruns. Carry on..

joostdevries:
:+1:

Gaurav0:
Fixed.

PR #59, opened 2015-07-17
Create an option to turn off autorun

Gaurav0:
Fixes issue #30

### DIFF
--- a/app/gist/controller.js
+++ b/app/gist/controller.js
@@ -65,7 +65,7 @@ export default Em.Controller.extend({
     }
   }),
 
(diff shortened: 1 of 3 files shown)


app/gist/controller.js
```diff
@@ -60,7 +66,9 @@ export default Em.Controller.extend({
   }),
 
   rebuildApp: Em.observer('model.files.@each.content', function() {
-    Em.run.debounce(this, this.buildApp, 500);
+    if (this.get('isAutorun')) {
```
joostdevries: you will also need to watch `isAutoRun` here I think so it runs when someone re-enables autorun.

PR #70, opened 2015-07-23
focus first column editor on file creation

miguelcobain:
`activeEditor` isn't a component instance anymore, but a column number.


joostdevries:
:+1:

PR #72, opened 2015-07-23
Fix bug in missing_gist

Gaurav0:
Oops!


miguelcobain:
One of those cases where using `==` could have prevented bad things :smile:

Gaurav0:
jshint does not like ==

PR #73, opened 2015-07-23
Use blueprints + cleanup build

joostdevries:


blueprints/twiddle.json
```diff
@@ -0,0 +1,8 @@
+{
+  "version": "0.3.5",
+  "dependencies": {
+    "jquery": "https://cdnjs.cloudflare.com/ajax/libs/jquery/1.9.1/jquery.js",
```
Gaurav0: 1.11.3

app/services/ember-cli.js
```diff
@@ -1,7 +1,22 @@
...
-import config from "../config/environment";
+
+const twiddleAppName = 'demo-app';
+
+// These files will be included if not present
+const boilerPlateJs = [
+  'app',
+  'config/environment',
+  'router',
+  'initializers/router'
```
Gaurav0: This should not be exposed to the user if possible.

app/services/ember-cli.js
```diff
@@ -90,16 +131,16 @@ export default Em.Service.extend({
...
-    var templateCode = Em.HTMLBars.precompile(code || '');
-    return this.compileJs('export default Ember.HTMLBars.template(' + templateCode + ');', moduleName);
+  compileHbs (code, filePath) {
+    let templateCode = Em.HTMLBars.precompile(code || '');
+    return this.compileJs('export default Ember.HTMLBars.template(' + templateCode + ');', filePath);
   },
 
   compileCss(code, moduleName) {
-    var prefix = "demo-app/styles/";
+    var prefix = "app/styles/";
```
Gaurav0: Why shouldn't this have the same moduleName as all other files?

app/services/ember-cli.js
```diff
@@ -20,58 +43,75 @@ export default Em.Service.extend({
...
-      // Add config
-      out.push(this.compileJs('export default {modulePrefix:"demo-app"}', 'demo-app/config/environment'));
+      this.checkRequiredFiles(out, gist);
+      this.addBoilerPlateFiles(out, gist);
 
       // Add boot code
-      contentForAppBoot(out, {modulePrefix:'demo-app'});
+      contentForAppBoot(out, {modulePrefix: twiddleAppName});
+
+      console.log(out.join('\n'));
```
Gaurav0: Remove log

app/services/ember-cli.js
joostdevries: if I read correctly the module name is actually obsolete right?

app/services/ember-cli.js
joostdevries: it's not, I just want the code to live in it's own file instead of being embedded inline.

PR #75, opened 2015-07-24
Adds CodeMirror font-family fallback to monospace

makepanic:
Monaco only ships with [OS X](https://en.wikipedia.org/wiki/Monaco_(typeface)) and is currently used as the CodeMirror editor font-family.
This commit adds a fallback to the default browser monospace font if Monaco isn't available.

fiddle-monospace


joostdevries:
:+1:

PR #76, opened 2015-07-24
Build a html string from services/ember-cli

joostdevries:
My idea is basically to have all of the build logic in `services/ember-cli` so that we don't have to do the same stuff in multiple places.. That way we can probably also unit-test `services/ember-cli` and have a pretty good baseline for overall functionality


app/gist/controller.js
```diff
@@ -13,7 +13,7 @@ export default Em.Controller.extend({
    * Output from the build, sets the `code` attr on the component
    * @type {String}
    */
-  buildOutput: Ember.Object.create({ code: '', styles: ''}),
+  buildOutput: '',
```
Gaurav0: What is the reason for changing this default?

joostdevries:
@Gaurav0 @rwjblue this further simplified the process, whatyathink?

rwjblue:
:+1: to the changes, would you mind adding a super simple unit test as a starting point for future testing initiatives?

Gaurav0:
@joostdevries While definitely simpler, I want to ensure that we don't expose index.html to the user until we know we've ensured he can't break the app (other than the build of his own app) with it.

joostdevries:
@Gaurav0 at this point it's not exposed as I'm taking it directly out of `blueprints` instead of merging the blueprint into `gist.files`

joostdevries:
@rwjblue test :-)

tests/unit/services/ember-cli-test.js
```diff
@@ -0,0 +1,42 @@
...
+  needs: ['model:gist','model:gistFile']
+});
+
+// Replace this with your real tests.
+test('compiling a gist works', function(assert) {
+  assert.expect(7);
+  var service = this.subject();
+  assert.ok(service);
+
+  var gist = Ember.Object.create({
```
joostdevries: using actual models here overcomplicated things, so I chose to use normal objects.

rwjblue: :+1:

tests/unit/services/ember-cli-test.js
```diff
@@ -0,0 +1,42 @@
...
+        extension: '.js',
+        content: 'import Ember from "ember";\n\nexport default Ember.Controller.extend({appName:"foo"});'
+      })
+    ])
+  });
+
+  Ember.run(function() {
+    service.compileGist(gist).then(function(output) {
+      output = output.replace(/define\('([a-z0-9\-\/]+)'/gi,'define("$1"');
+      console.log(output);
```
rwjblue: extra `console.log`

rwjblue:
:+1: - Looks good, left one small comment...

Issue #79, opened 2015-07-24
[BUG] Cannot add two files of the same type, even if name is different.


Gaurav0:
Steps to reproduce:

Create new twiddle.
Add template, name it 'foo/template.hbs'
Try to add another template.

You are not prompted to change the name. In the console, you see the error "Uncaught Error: Assertion Failed: The id my-route.template.hbs has already been used with another record of type ember-twiddle@model:gist-file:".

Gaurav0:
For now, this fix is deployed on https://ember-twiddle-gaurav.herokuapp.com

rwjblue:
Fixed by #80.

PR #80, opened 2015-07-24
Fix issue cannot create two files of the same type, even if the name is different

Gaurav0:
Fixes issue #79


stefanpenner:
is it possible to add a regression test?

Gaurav0:
@stefanpenner Yes, done.

stefanpenner:
wunderbar

app/adapters/gist-file.js
```diff
@@ -1,7 +1,9 @@
 import ApplicationAdapter from './application';
 
 export default ApplicationAdapter.extend({
+  seq: 0,
+
 	generateIdForRecord: function(store, type, inputProperties) {
-	  return inputProperties.filePath.replace(/\//gi, '.');
+	  return inputProperties.filePath.replace(/\//gi, '.') + "." + this.incrementProperty('seq');
```
rwjblue: Instead of allowing duplicate filePath's, I think that we should delay calling `store.createRecord` until after we have prompted for the file name. 

We should:

* Add a `build` (or `buildProperties`) to `ember-cli` service.
* Move main logic out of `generate` into `build`, leaving `generate` as roughly `return this.store.createRecord('gistFile', this.build(type));`.
* Update `addNew` action in `gist/controller.js` to call `build`, then prompt, then createRecord.

Gaurav0:
@rwjblue How's this?

Gaurav0:
Can't seem to get Travis to test this.

rwjblue:
Looks good to me, thanks for working on this!

Issue #81, opened 2015-07-26
Still cannot run older version of ember.js


Gaurav0:
See https://ember-twiddle-gaurav.herokuapp.com/e99546e0831d6d0251fc

Possibly loading / using the wrong version of ember-template-compiler?

rwjblue:
We need to add the template compiler to `twiddle.json`.  We can use the default one from https://cdnjs.cloudflare.com/ajax/libs/ember.js/1.13.5/ember-template-compiler.js

Gaurav0:
@rwjblue Thanks. The only issues now is that 

1) we no longer precompile templates
2) This breaks existing twiddles

Gaurav0:
Fixed.

PR #83, opened 2015-07-26
Use external ember-template-compiler

Gaurav0:
Use external ember-template-compiler to allow for older versions of Ember.js

Fixes Issue #81

### DIFF
--- a/app/services/ember-cli.js
+++ b/app/services/ember-cli.js
@@ -210,8 +210,11 @@ export default Em.Service.extend({
    * @return {String}            AMD module code
    */
   compileHbs (code, filePath) {
-    let templateCode = Em.HTMLBars.precompile(code || '');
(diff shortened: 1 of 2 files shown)


rwjblue:
Submitted https://github.com/Gaurav0/ember-twiddle/pull/1 to this branch, which should allow existing twiddle's to function properly.

rwjblue:
We should add a test to confirm that various versions of Ember work (boot the app in the iframe), we can also add a variant that does not include `ember-template-compiler` in the `twiddle.json` to ensure existing twiddles work properly.

Gaurav0:
@Thanks rwjblue. I'm generalizing your work a bit.

Gaurav0:
This is complete and ready for review.

rwjblue:
Thank you!

Issue #87, opened 2015-07-27
No confirmation prompt when deleting twiddle


abidmix:
A confirmation prompt should pop up when deleting a tweedle to avoid accidental deletions

rwjblue:
:+1: - Agreed!

rwjblue:
Fixed by #90.

PR #85, opened 2015-07-26
Write a routing acceptance test

Gaurav0:
This adds an acceptance test using ember-cli-mirage and a whole bunch of interesting hacks.

PR #90, opened 2015-07-27
Confirm before deleting a twiddle

Gaurav0:
Fixes issue #87

### DIFF
--- a/tests/unit/gist/controller-test.js
+++ b/tests/unit/gist/controller-test.js
@@ -2,19 +2,18 @@ import Ember from "ember";
 import { moduleFor, test } from 'ember-qunit';
 
 moduleFor('controller:gist', {
-  // Specify the other units that are required for this test.
-  // needs: ['controller:foo']
-  needs: ['service:ember-cli']
-});
+  needs: ['service:ember-cli'],
 
-// Replace this with your real tests.
-test('it exists', function(assert) {
-  var controller = this.subject();
-  assert.ok(controller);
+  beforeEach() {
+    this._originalConfirm = window.confirm;
+  },
+
+  afterEach() {
+    window.confirm = this._originalConfirm;
+  }
 });
 
 test('deleting a gist requires confirmation', function(assert) {
-  let cacheConfirm = window.confirm;
   let controller = this.subject({
     transitionToRoute() {},
     notify: {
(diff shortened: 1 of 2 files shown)


rwjblue:
Can you add a test for this?

A unit test will likely be easiest (you could mock `confirm` and pass in an object with a `destroyRecord` method then assert it is called when `confirm` returns true and not called when `confirm` returns false), but an acceptance test would be better.  Acceptance test would be harder, but still should be doable (similar to the changes made in #85).

Gaurav0:
@rwjblue Unit test added.

tests/unit/gist/controller-test.js
```diff
@@ -0,0 +1,42 @@
...
+});
+
+// Replace this with your real tests.
+test('it exists', function(assert) {
+  var controller = this.subject();
+  assert.ok(controller);
+});
+
+test('deleting a gist requires confirmation', function(assert) {
+  let cacheConfirm = window.confirm;
```
rwjblue: We should grab this in the `beforeEach` and reset it in the `afterEach`.  The main reason is that if an error is thrown during the test `window.confirm` will not be reset (and will likely cause other tests to fail in the future).

Something like:

```javascript
moduleFor('controller:gist', {
  needs: ['service:ember-cli'],
  beforeEach() {
    this._originalConfirm = window.confirm;
  },

  afterEach() {
    window.confirm = this._originalConfirm;
  }
});
```

tests/unit/gist/controller-test.js
```diff
@@ -0,0 +1,42 @@
+import Ember from "ember";
+import { moduleFor, test } from 'ember-qunit';
+
+moduleFor('controller:gist', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
```
rwjblue: Can you delete this comment?

tests/unit/gist/controller-test.js
```diff
@@ -0,0 +1,42 @@
...
+  needs: ['service:ember-cli']
+});
+
+// Replace this with your real tests.
```
rwjblue: This test doesn't add value now that you added a real one below, can you delete it?

rwjblue:
@Gaurav0 - Awesome, thank you!  I left a few small comments, but this is looking good!

Gaurav0:
@rwjblue Comments addressed.

rwjblue:
:+1: - Thank you!

abidmix:
Thanks for fixing this prompty.

Issue #84, opened 2015-07-26
Create a test helper to run an acceptance test against a gist


Gaurav0:
Make it easier to write an acceptance test tha "loads a gist" by mocking the github api responses.

rwjblue:
I'd like to see us have a few more acceptance test helpers to make adding tests much easier.

Off the top of my head, I think the following would be a great start:

* `addFile` - Takes a type, and filename, then adds a file via the normal UI.
* `deleteFile` - Takes the filename, and removes it via the normal UI.
* `changeDisplayedFile` - Takes the filename to display and either `"left"` or `"right"`, then uses the normal UI to change the file displayed in panes.
* `runGist` - Accepts a stubbed gist, and runs the app.

Many of these can be extracted from the recent acceptance tests added by @Gaurav0.

PR #94, opened 2015-07-28
Create runGist helper; minimal gist acceptance test

Gaurav0:
Relevant to issue #84

### DIFF
--- a/tests/acceptance/routing-test.js
+++ b/tests/acceptance/routing-test.js
@@ -38,7 +38,7 @@ test('Able to do routing in a gist', function(assert) {
       content: "{\n  \"version\": \"0.4.0\",\n  \"dependencies\": {\n    \"jquery\": \"https://cdnjs.cloudflare.com/ajax/libs/jquery/1.11.3/jquery.js\",\n    \"ember\": \"https://cdnjs.cloudflare.com/ajax/libs/ember.js/1.13.5/ember.js\",\n    \"ember-data\": \"https://cdnjs.cloudflare.com/ajax/libs/ember-data.js/1.13.5/ember-data.js\"\n  }\n}"
     }
   ];
-  
+
   const aboutLink = '.test-about-link';
   const indexLink = '.test-index-link';
   const outletText = 'p';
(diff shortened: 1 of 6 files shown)


joostdevries:
I actually like this a lot :-)

joostdevries:
@rwjblue what do you think?

tests/helpers/start-app.js
```diff
@@ -10,6 +11,7 @@ export default function startApp(attrs) {
 
   Ember.run(function() {
     application = Application.create(attributes);
+    Ember.Test.registerAsyncHelper('runGist', runGist);
```
rwjblue: This only needs to happen once, when the module is loaded.  Including it here (inside the `startApp` function) will have this ran once for each acceptance test.

I prefer to have the `run-gist` helper imported and registered in `tests/test-helper.js`.  Any objections?

tests/acceptance/routing-test.js
```diff
@@ -15,87 +15,51 @@ module('Acceptance | routing', {
...
-        return resolve();
-      });
-    });
+    iframe_window.click(iframe_window.find(aboutLink));
   });
 
   andThen(function() {
-    iframe_window.visit('/');
-  });
+    assert.equal(iframe_window.find(outletText).text().trim(), 'About Page', 'About Link leads to About Page being displayed');
```
rwjblue: This is repeated a few times, can it be extracted to a test helper?

Something like:

```javascript
const iframe = '#dummy-content-iframe';

Ember.Test.registerHelper('outputPane', function(app) {
  return find(iframe)[0].contentWindow;
});

Ember.Test.registerHelper('outputContents', function(app, selector) {
  let output = app.testHelpers.outputPane();
  return output.find(selector).text().trim();
});

// then you can use them as follows:
outputPane().click(aboutLink);
andThen(() => {
  assert.equal(outputContents(outletText), 'About Page');
});
```

Gaurav0:
@joostdevries @rwjblue Updated with suggested changes.

rwjblue:
Awesome!  Thank you!

PR #95, opened 2015-07-28
Implement Unsaved Indicator

Gaurav0:


app/gist/controller.js
```diff
@@ -74,6 +75,11 @@ export default Em.Controller.extend({
   }),
 
   rebuildApp: Em.observer('model.files.@each.content', 'isAutorun', function() {
+    if (!this.get('unsaved')) {
+      Em.run.scheduleOnce('sync', function() {
+        this.set('unsaved', true);
```
stefanpenner: It seems like this could just be a computed property

app/gist/controller.js
```diff
@@ -74,6 +75,11 @@ export default Em.Controller.extend({
   }),
 
   rebuildApp: Em.observer('model.files.@each.content', 'isAutorun', function() {
+    if (!this.get('unsaved')) {
+      Em.run.scheduleOnce('sync', function() {
```
rwjblue: `Ember.run.scheduleOnce` does not really work with anonymous functions (because it uses `func1 === func2` to know if it should schedule again, and in the anon function case they are always different).  Can probably do `Ember.run(() => this.set('unsaved', true));` instead.

app/gist/controller.js
```diff
@@ -15,6 +15,7 @@ export default Em.Controller.extend({
    */
   buildOutput: Ember.Object.create({ code: '', styles: ''}),
   isBuilding: false,
+  unsaved: true,
```
rwjblue: Can we make this a CP on the underlying `model`?  If we can, then the various `setupController` functions implemented here can be removed...

joostdevries:
Wouldn't something like https://github.com/lolmaus/ember-cli-stained-by-children solve this?

Gaurav0:
@stefanpenner @rwjblue I must be missing something obvious. I can't figure out how to make unsaved a computed property.

Gaurav0:
@rwjblue @stefanpenner @joostdevries Refactored. Please review again.

app/components/run-or-live-reload.js
```diff
@@ -0,0 +1,21 @@
...
+export default Ember.Component.extend({
+  classNames: ['run-or-live-reload'],
+
+  /**
+   * Whether user wishes the code to automatically run
+   * @type {boolean}
+   */
+  isLiveReload: true,
+
+  liveReloadDidChange: Ember.observer('isLiveReload', function() {
```
stefanpenner: can we use an action from the template instead of this observer?

app/components/file-editor-column.js
```diff
@@ -19,6 +21,10 @@ export default Em.Component.extend({
     this.sendAction('focusEditor', this);
   },
 
+  contentsDidChange: function() {
+    this.sendAction('contentsChanged');
+  }.observes('file.content'),
```
stefanpenner: can we get an event/action instead of observing here as well?

joostdevries:
[there are no observers](http://m.youtube.com/watch?v=7PUX27RKCq0) ;-)

app/components/run-or-live-reload.js
Gaurav0: Nope. Not until Ember 2.0. https://github.com/emberjs/ember.js/issues/5433

app/components/file-editor-column.js
Gaurav0: ivy-codemirror only provided a binding, not an action. Sorry.

Gaurav0:
@joostdevries I wish.

Gaurav0:
@stefanpenner @rwjblue Still waiting for a hint as to how all this could have been done with a computed property.

app/components/run-or-live-reload.js
joostdevries: doesn't this work:
```handlebars
<input type="checkbox" value={{isChecked}} onclick={{action "foo" value="target.checked"}} />
```

Gaurav0: No.

Gaurav0: Found a solution. We can write our own checkbox component.

stefanpenner:
I'm confused why @joostdevries example doesn't work for this case

Gaurav0:
@stefanpenner I really don't know. I just tried it and it didn't work.

Gaurav0:
@joostdevries @stefanpenner @rwjblue Observers removed. Contortions added. Please review again.

app/gist/controller.js
```diff
@@ -130,6 +141,9 @@ export default Ember.Controller.extend({
         this.notify.info('File %@ was added'.fmt(file.get('filePath')));
         this.set('col1File', file);
         this.set('activeEditorCol', '1');
+
+        this.set('unsaved', true);
+        this.rebuildApp();
```
joostdevries: why not just send `contentsChanged` here?

app/gist/controller.js
```diff
@@ -164,6 +178,9 @@ export default Ember.Controller.extend({
         file.deleteRecord();
         this.notify.info('File %@ was deleted'.fmt(file.get('filePath')));
         this._removeFileFromColumns(file);
+
+        this.set('unsaved', true);
+        this.rebuildApp();
```
joostdevries: why not just send `contentsChanged` here?

app/gist/edit/route.js
```diff
@@ -7,6 +8,16 @@ export default GistRoute.extend({
...
+  setupController() {
+    this._super.apply(this, arguments);
+
+    let gistController = this.controllerFor('gist');
+    gistController.rebuildApp();
+    Ember.run.schedule('afterRender', function() {
+      gistController.set('unsaved', false);
+    });
+  },
+
```
joostdevries: Why is this code needed?

app/gist/controller.js
Gaurav0: Ok. will do.

app/gist/controller.js
Gaurav0: Ok. Will do.

app/gist/edit/route.js
Gaurav0: So the observer was synchronous, while the action waits for it's turn in the 'actions' queue. We need to set 'unsaved' to false after the action has taken place.

Gaurav0:
@joostdevries Updated.

joostdevries:
:+1:

PR #98, opened 2015-07-29
More acceptance tests!

Gaurav0:


tests/acceptance/dependency-test.js
```diff
@@ -0,0 +1,51 @@
...
+
+let application;
+
+module('Acceptance | external dependency', {
+  beforeEach: function() {
+    application = startApp();
+  },
+
+  afterEach: function() {
+    Ember.run(application, 'destroy');
```
stefanpenner:
```
this.application = ..

Ember.run(this.application)
```

should help mitigate some leaks :)

rwjblue: @Gaurav0 - To clarify, @stefanpenner is referring to avoiding usage of the `var application` variable here.  Save the created instance on `this` instead (helps the test framework ensure proper cleanup is done).

app/services/ember-cli.js
```diff
@@ -80,7 +80,8 @@ const availableBlueprints = {
 const requiredDependencies = [
   'jquery',
   'ember',
-  'ember-template-compiler'
+  'ember-template-compiler',
+  'ember-data'
```
rwjblue: I am slightly concerned with making `ember-data` required.  It is probably fine, but should be considered independently...

rwjblue: While I think ember-data is awesome (and use it often), I also find it useful for demonstration purposes to be able to use JSBin without Ember Data.  Changing this will make it impossible to create a twiddle without Ember Data.

Perhaps, we can still add here and use a `null` value in `twiddle.json` when we want to explicitly not include it?  Not sure what the best path forward here is....

tests/acceptance/version-test.js
```diff
@@ -0,0 +1,46 @@
...
+  beforeEach: function() {
+    application = startApp();
+  },
+
+  afterEach: function() {
+    Ember.run(application, 'destroy');
+  }
+});
+
+test('Able to load a gist without a template compiler', function(assert) {
```
rwjblue: This seems to be an incorrect test description.

rwjblue:
LOVE THE TESTS!!! THANK YOU!!!

app/services/ember-cli.js
Gaurav0: I added it because the test failed if I didn't.

Gaurav0: Also, if a user left off ember-data, a gist is currently not loading. This is a bug.

Gaurav0:
@rwjblue @stefanpenner Updated as per your comments.

app/services/ember-cli.js
Gaurav0: Never mind, I fixed the bug. Ember Data no longer required.

PR #103, opened 2015-07-30
Use sandbox + srcdoc and postMessage for updating the url.

### LINKED ISSUE #99: Should the iframe point to a different domain?
Firstly, the app is really cool. Hats off!!

In order to prevent XSS attacks and other security issues, should the iframe in ember-twiddle point to a different domain? Similar to JSFiddle. Currently, we can access the parent window from the iframe code and can access the browser cookies as well.  

Correct me if my understanding is wrong.

joostdevries:
Fixes #99. THIS WIL BREAK IE.

- [x] Implement sandbox
- [x] Fix broken tests
- [ ] Add test for sandbox
- [x] Add blocking message for IE / unsupported browser users

### DIFF
--- a/app/gist/controller.js
+++ b/app/gist/controller.js
@@ -1,7 +1,8 @@
+import Ember from "ember";
 import config from '../config/environment';
 
-export default Em.Controller.extend({
-  emberCli: Em.inject.service('ember-cli'),
+export default Ember.Controller.extend({
+  emberCli: Ember.inject.service('ember-cli'),
   version: config.APP.version,
   init() {
     this._super(...arguments);
@@ -60,7 +61,7 @@ export default Em.Controller.extend({
   /**
    * Set the initial file columns
    */
-  initializeColumns: Em.observer('model', function() {
+  initializeColumns: Ember.observer('model', function() {
     var files = this.get('model.files');
 
     if(files.objectAt(0)) {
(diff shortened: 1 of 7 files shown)


joostdevries:
@Gaurav0 this is quite tricky since the tests need access to the iframe. I could disable the sandbox during testing but that would kind of defeat the point of the test right?

Gaurav0:
The more I think about it the more I think we need to go with the mutlple domains.

joostdevries:
You will have the same problem.

Gaurav0:
I guess I could accept a toggle based on testing as long as there was no use of window.parent from the iframe.

Gaurav0:
Can you enforce that with a window.parent = window?

app/gist/controller.js
```diff
@@ -50,6 +49,7 @@ export default Em.Controller.extend({
       this.set('buildOutput', buildOutput);
     })
     .catch(errors => {
+      console.log(errors);
```
Gaurav0: Do you need this still?

joostdevries: sorry no.. will remove.

joostdevries:
We need a nice error message for IE users but the benefit of srcdoc is that it will just not work (so we don't have to block manually).

joostdevries:
Also https://github.com/jugglinmike/srcdoc-polyfill

Gaurav0:
Is the polyfill secure?

joostdevries:
would like to merge as-is and release, if someone could signoff that'd be great.

joostdevries:
image

stefanpenner:
Polyfil looks safe and legit. some browser do have a uri limit (bug) but I think it's only safari that hits the issue with relatively low size. 4k. And safari supports real srcdoc, so all seems well based on my current recollection 

+1

stefanpenner:
This looks good as is. The poly can come now or later.

PR #109, opened 2015-07-31
Reenable a content security policy (very loose)

Gaurav0:


rwjblue:
This mostly affects development only, we need to add additional headers to be returned by the server when our assets are requested for it to take effect when deployed.  Also, it seems that AWS does not support the `Content-Security-Policy` header for S3 objects (see [here](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html)).

joostdevries:
I'd close this for now since AWS doesn't support it.

Gaurav0:
Closed due to lack of AWS support.

PR #115, opened 2015-07-31
Add .watchmanconfig

HeroicEric:
Adds `.watchmanconfig` that now comes with ember-cli

https://github.com/ember-cli/ember-cli/pull/4124


rwjblue:
Thanks!

PR #117, opened 2015-08-01
Remove usage of global `Em`

HeroicEric:
Adds missing imports of `Ember` and removes usage of global `Em`

### DIFF
--- a/app/models/gist-file.js
+++ b/app/models/gist-file.js
@@ -1,16 +1,20 @@
 import DS from 'ember-data';
+import Ember from 'ember';
 import Path from 'npm:path';
 
+const { attr, belongsTo } = DS;
+const { computed, observer } = Ember;
+
 export default DS.Model.extend({
-  fileType: DS.attr('string'),
-  fileName: DS.attr('string'),
-  content: DS.attr('string'),
-  gist: DS.belongsTo('gist', {async:false}),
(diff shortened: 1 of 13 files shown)


rwjblue:
Can you remove from [.jshintrc](https://github.com/ember-cli/ember-twiddle/blob/master/.jshintrc#L7) also?

joostdevries:
@HeroicEric could you merge master/rebase.

HeroicEric:
@joostdevries updated

app/models/gist-file.js
```diff
@@ -10,7 +11,7 @@ export default DS.Model.extend({
   /*
     Replace dots with slashes, Gists can't have directories
    */
-  filePath: Em.computed('fileName', function(key, value) {
```
stefanpenner: can we instead do

```js
const { computed } = Ember
```

this will be nice once we cut over to the module API.

HeroicEric: For sure. Didn't want to start making style decisions but I'm :+1: if you all are

stefanpenner: :+1: from me

HeroicEric: @stefanpenner Updated that one and a few more that I came across

stefanpenner:
nice thanks!

PR #119, opened 2015-08-01
Added hand cursor style for all the menu options

phkavitha:
Currently, in the application most of the menu options shows text cursor on hover which does not give us a feeling that it's clickable. 

In this pull request, I have added the hand cursor style for all the drop-down menu options so that it's consistent across the application.


Gaurav0:
Looks good to me. :+1:

Issue #127, opened 2015-08-03
Refactor gist/controller.js and gist/template.hbs into separate components for each ui element


Gaurav0:
So the gist/controller.js and particularly gist/template.hbs have gotten a bit large. It is time to separate this out into components.

Gaurav0:
Fixed by #128, #130, #137, #138

PR #128, opened 2015-08-04
Extract File menu into a component

Gaurav0:
Relevant to issue #127

### DIFF
--- a/tests/integration/components/file-menu-test.js
+++ b/tests/integration/components/file-menu-test.js
@@ -2,40 +2,38 @@ import Ember from "ember";
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
 
-let file, model;
-
-let addFileCalled = false;
-let renameFileCalled = false;
-let removeFileCalled = false;
-let saveGistCalled = false;
-let forkCalled = false;
-let deleteGistCalled = false;
-let signInViaGithubCalled = false;
-
-let fileType = null;
-let renamedFile = null;
-let removedFile = null;
-let gistToSave = null;
-let gistToFork = null;
-let gistToDelete = null;
-
 moduleForComponent('file-menu', 'Integration | Component | file menu', {
   integration: true,
   beforeEach() {
 
-    // Set any properties with this.set('myProperty', 'value');
-    file = Ember.Object.create({
+    this.addFileCalled = false;
+    this.renameFileCalled = false;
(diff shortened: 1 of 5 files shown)


rwjblue:
Thank you, I definitely like this!  A few tweaks:

* Can you swap the `internalAction="someActionFromControlleR"` and `this.sendAction` usage for the newer style closure actions?
* Can you split up the test a bit (I'd like to see a separate test for each action)?

Gaurav0:
@rwjblue Updated. Some of the actions could not be made into closure actions because they bubble to the route.

tests/integration/components/file-menu-test.js
```diff
@@ -0,0 +1,132 @@
...
+    this.set('session', Ember.Object.create({
+      isAuthenticated: true
+    }));
+
+    this.set('activeEditorCol', 1);
+
+    this.set('activeFile', file);
+
+    // Handle any actions with this.on('myAction', function(val) { ... });
+    this.on('addFile', (type) => { addFileCalled = true; fileType = type; });
```
rwjblue: Can you tweak these to save to `this`?  Something like:

```javascript
this.on('addFile', (type) => { this.addFileCalled = true; this.fileType = type; });
```

Then you can assert in the tests below like:

```javascript
assert.ok(this.addFileCalled, 'addFile was called');
assert.equal(this.fileType, 'template', 'addFile was called with type parameter');
```

---

Using the above, will help us ensure that there is no leakage between tests.  Say we add a test in the future that also triggers an action, the boolean `xyzCalled` may have already been set causing the tests to be not quite right.

Gaurav0:
@rwjblue Updated again as per request.

tests/integration/components/file-menu-test.js
```diff
@@ -0,0 +1,130 @@
+import Ember from "ember";
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('file-menu', 'Integration | Component | file menu', {
+  integration: true,
+  beforeEach() {
+
+    this.addFileCalled = false;
```
rwjblue: This is totally fine as is, but removing these would result in the same test (since `this.addFileCalled` would be undefined and not truthy the assertions would still pass).

rwjblue:
:+1: -   Sorry to be such an annoying review, but thanks for working so hard on this!

PR #129, opened 2015-08-04
Quick fix for cutoff title

Gaurav0:
This fixes the following issue on canary:

screen shot 2015-08-04 at 4 36 06 pm

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -14,8 +14,8 @@ $topbar-control-height: 32px;
 $code-header-height: 60px;
 
 .dropdown-submenu:hover > a {
-	background: $dropdown-link-hover-bg;
-	color: $dropdown-link-hover-color;
+  background: $dropdown-link-hover-bg;
+  color: $dropdown-link-hover-color;
 }
 
 @import "bower_components/bootstrap-sass/assets/stylesheets/bootstrap";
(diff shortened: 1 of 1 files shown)


miguelcobain:
This always annoyed me.

:+1:

joostdevries:
does not fix it for me on dev

app/styles/app.scss
```diff
@@ -98,6 +98,7 @@ body {
 		  border: 0;
 		  color: #faf2ee;
 			cursor: pointer;
+      box-sizing: content-box;
```
joostdevries: you need to apply this on the `input` element itself, not the wrapper.

Gaurav0:
@joostdevries Seems that somewhere you have a * { box-sizing: border-box } on canary that I don't have on my local copy of master. So it inherited for me, but not for you. Why don't we have the same styles?

Anyway, fixed.

app/styles/app.scss
```diff
@@ -103,6 +103,7 @@ body {
 				background-color: transparent;
 				border: 0;
 				outline: 0;
+        box-sizing: content-box;
```
joostdevries: what's up with the indent?

Gaurav0: tabs vs spaces. the usual.

joostdevries: maybe convert the entire file to spaces to avoid messiness.

Gaurav0: Done.

joostdevries:
:+1:

PR #131, opened 2015-08-05
Remove GitHub fork ribbon

jpadilla:
After @miguelcobain's great work updating styles I think we could get rid of the "Fork me on GitHub" ribbon. IMHO it's way too big and distracting. We're already linking to the repo and issue tracker from the top menu. Thoughts?


rwjblue:
I'm in favor of removing it.

miguelcobain:
:+1:

jpadilla:
Great, thanks!

Issue #133, opened 2015-08-05
"Fork Twiddle" should fork, then save changes, then transition


Gaurav0:
So that if the user has made any changes to the code, they are not lost.

nahtnam:
:+1:

Gaurav0:
Fixed by PR #134

PR #134, opened 2015-08-05
Fork twiddle should save changes

Gaurav0:
Fixes issue #133

### DIFF
--- a/app/gist/route.js
+++ b/app/gist/route.js
@@ -38,8 +38,10 @@ export default Ember.Route.extend({
           let firstError = error.errors[0];
           if (firstError.code === "unprocessable" && firstError.field === "forks") {
             this.notify.info("You already own this gist.");
+            return;
           }
         }
+        throw error;
       });
     },
 
--- a/app/gist/route.js
+++ b/app/gist/route.js
@@ -28,11 +28,18 @@ export default Ember.Route.extend({
 
     fork (gist) {
       gist.fork().then((response) => {
-        var id = response.id;
-        gist.set('id', id);
-        return gist.save().then(() => {
-          this.transitionToRoute('gist.edit', id);
+        let id = response.id;
+        let newGist = this.duplicate(gist);
+        return newGist.save().then(() => {
+          this.transitionTo('gist.edit', id);
         });
+      }).catch((error) => {
+        if (error && error.errors) {
+          let firstError = error.errors[0];
+          if (firstError.code === "unprocessable" && firstError.field === "forks") {
+            this.notify.info("You already own this gist.");
+          }
+        }
       });
     },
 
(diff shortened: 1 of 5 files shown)


stefanpenner:
LGTM

Gaurav0:
Wait, got a bug.

Gaurav0:
@stefanpenner OK, now you can review.

app/gist/route.js
```diff
@@ -36,5 +53,20 @@ export default Ember.Route.extend({
...
+    Object.keys(newGistJson).forEach((key) => {
+      if (newGistJson[key] !== gist.get(key)) {
+        newGistJson[key] = gist.get(key);
+      }
+    });
+    newGistJson.id = undefined;
+    var newGist = this.get('store').createRecord('gist', newGistJson);
+    gist.get('files').currentState.forEach((file) => {
+      file.record.set('gist', newGist);
+    });
```
Gaurav0: What is the correct way to do this?

app/gist/route.js
```diff
@@ -26,6 +26,23 @@ export default Ember.Route.extend({
...
+        return newGist.save().then(() => {
+          this.transitionTo('gist.edit', id);
+        });
+      }).catch((error) => {
+        if (error && error.errors) {
+          let firstError = error.errors[0];
+          if (firstError.code === "unprocessable" && firstError.field === "forks") {
+            this.notify.info("You already own this gist.");
+          }
+        }
```
stefanpenner: if we did not understand the error, we should re-throw

stefanpenner:
there appears to be very API related code in the route, it might be nice to extract that.

Gaurav0:
Hey, the ‘very api related code’ uses private API. I was wondering if there was a correct, public, way to do that.

app/gist/route.js
stefanpenner: since this is async:false `toArray()` should work.

app/gist/route.js
Gaurav0: Done.

app/gist/route.js
```diff
@@ -26,6 +26,25 @@ export default Ember.Route.extend({
...
+    fork (gist) {
+      gist.fork().then((response) => {
+        let id = response.id;
+        let newGist = this.duplicate(gist);
+        return newGist.save().then(() => {
+          this.transitionTo('gist.edit', id);
+        });
+      }).catch((error) => {
+        if (error && error.errors) {
+          let firstError = error.errors[0];
```
stefanpenner: extracting the error callback into its own function, so it can be united tested will help cleanup this code some more. The  extra verbosity distracts from the overall flow of data.

Gaurav0:
Again, it wasn't working correctly, so I took another approach.

Gaurav0:
@stefanpenner Updated. Please review again.

mmun:
Ideally you would keep the closure actions and move the actions off the router and into a service.

Gaurav0:
@mmun Perhaps, but that should be a separate PR. Can you open an issue?

Gaurav0:
@joostdevries @stefanpenner @rwjblue Can I get a review?

stefanpenner:
LGTM

PR #137, opened 2015-08-08
Extract build-messages component

Gaurav0:


rwjblue:
Looks great!  Can you add some cursory integration tests for the new component?

Gaurav0:
@rwjblue Just did.

tests/integration/components/build-messages-test.js
```diff
@@ -0,0 +1,50 @@
...
+  this.set('isBuilding', false);
+  this.set('notify', notifyObject);
+
+  this.render(hbs`{{build-messages buildErrors=buildErrors isBuilding=isBuilding notify=notify}}`);
+
+  this.$('span a').click();
+
+  assert.ok(notifyObject.get('called'), "notify.info() was called");
+
+  console.error = this.cacheConsole;
```
rwjblue: This cleanup needs to be in the `afterEach` (in case the test fails we don't want to leave it in an invalid state).

tests/integration/components/build-messages-test.js
```diff
@@ -0,0 +1,50 @@
...
+  assert.expect(1);
+
+  let notifyObject = Ember.Object.create({
+    called: false,
+    info() {
+      this.set('called', true);
+    }
+  });
+
+  this.cacheConsole = console.error;
```
rwjblue: This should be in the `beforeEach` so that when we cleanup in the `afterEach` we will always have it available.

Gaurav0:
@rwjblue Comments addressed.

rwjblue:
@Gaurav0 - Awesome, thanks for the tests (you must have pushed right when I was commenting, sorry about that)!

Gaurav0:
@rwjblue No problem. The component wasn't working quite right too.

PR #140, opened 2015-08-08
Replace broccoli-sass with ember-cli-sass

HeroicEric:
Replaces broccoli-sass with [ember-cli-sass](https://github.com/aexmachina/ember-cli-sass).

ember-cli-sass is now the recommended addon for using sass http://www.ember-cli.com/user-guide/#sass


Gaurav0:
:+1:

Issue #122, opened 2015-08-01
Add support for more file editor panes


HeroicEric:
One thing that keeps me preferring JSBin right now is that it's hard to get the full context of what's going on by jumping around between files in ember twiddle. I think this can be improved a lot if we could add support for more file editor panes.

I think something that allows you to add/remove columns and horizontally split a column to show files above/below each other within a column would go a long way.

I'd like to try implementing this using [flexbox](https://philipwalton.github.io/solved-by-flexbox/demos/grids/) unless anybody has objections. [Browser support for flexbox](http://caniuse.com/#search=flexbox)

Looking for feedback

Gaurav0:
@joostdevries has specific ideas about the user interface, so I'll defer to him, but my feeling is that users should be allowed to do this if they have screen space, but if they follow a link that opens 6 files and they are on a small screen or mobile they may have some difficulty. Not sure what the solution is to that specific issue.

joostdevries:
I think a toggle with the # of editor cols (eg. 1-5) + a toggle whether or not to display output would pretty much do the trick. A horizontal split would clutter the UI I think (because of the controls you need for every editor).

For mobile, I'd say we'd some way to elegantly handle any possible amount of panes. Maybe hor. scroll or something..

HeroicEric:
Closing since this has been implemented by https://github.com/ember-cli/ember-twiddle/pull/141 and more

Issue #65, opened 2015-07-22
Resizeable / Hideable Columns


Gaurav0:

Gaurav0:
Fixed by #141, #170

PR #141, opened 2015-08-09
Hide and show columns, allow for zero to 3 columns, query param for numColumns

Gaurav0:
Relevant to #65, #100, #122


app/gist/controller.js
```diff
@@ -68,16 +100,21 @@ export default Ember.Controller.extend({
   /**
    * Set the initial file columns
    */
-  initializeColumns: observer('model', function() {
-    var files = this.get('model.files');
+  initializeColumns: observer('model', 'numColumns', function() {
```
HeroicEric: Might be nice to refactor this to use a computed property rather than an observer

app/gist/controller.js
```diff
@@ -68,16 +100,21 @@ export default Ember.Controller.extend({
...
+    let files = this.get('model.files');
+    let numColumns = Math.min(this.get('numColumns'), MAX_COLUMNS);
 
-    if(files.objectAt(0)) {
-      this.set('col1File', files.objectAt(0));
+    let j = 0;
+    for (let i = 0; i < numColumns; ++i) {
```
stefanpenner: i would just make a single array that is a list of columns, and make the creation of that array be in a CP

Gaurav0: Updated to remove observer.

Gaurav0:
This PR is complete. Please review.

Gaurav0:
@stefanpenner @joostdevries @rwjblue Updated. Pretty please review.

app/gist/controller.js
```diff
@@ -219,6 +291,28 @@ export default Ember.Controller.extend({
       }
     },
 
+    removeColumn (col) {
+      let numColumns = this.get('realNumColumns');
```
stefanpenner: I'm curious, it seems like rather then making this controller have array like qualities. Having the columns represented off-object would be clean. Something like:

```js
controller.columns === [
  col1,
  col2,
  col3
]
```

```js
columns: computed('dataColumnsNeed', function() {
  return ensureColums(this.get('dataColumnsNeed'))
});
```

the return columns object itself doesn't need to be an array, but may prove to be nice if it ways.

This would also make it easy to support N columns, if we are so inclined, but also begins to move to a pull based system over the imperative model.

app/gist/controller.js
```diff
@@ -219,6 +291,28 @@ export default Ember.Controller.extend({
...
+        this.set('activeEditorCol', ((activeCol|0) - 1).toString());
+      }
+
+      this.transitionToRoute({queryParams: {numColumns: numColumns - 1}});
+    },
+
+    addColumn() {
+      let numColumns = this.get('realNumColumns');
+
+      this.transitionToRoute({queryParams: {numColumns: numColumns + 1}}).then(this.initializeColumns.bind(this));
```
stefanpenner: multi line please

stefanpenner: > this.initializeColumns.bind(this)

feels like it should merely be invalidating something, maybe changing columnCount. The explicit `initialColumns` call enforces eager production of something that seems to really want to be pull based.

app/gist/controller.js
```diff
@@ -227,11 +321,10 @@ export default Ember.Controller.extend({
   },
 
   _removeFileFromColumns (file) {
-    if(this.get('col1File') === file) {
-      this.setColumnFile(1, null);
-    }
-    if(this.get('col2File') === file) {
-      this.setColumnFile(2, null);
+    for (let i = 1; i <= MAX_COLUMNS; ++i) {
```
stefanpenner: if columns becomes off object, this would become

```js
this.get('columns').removeObject(file)
```

app/gist/controller.js
```diff
@@ -123,6 +189,11 @@ export default Ember.Controller.extend({
     }
     return false;
   },
+
+  getColumnFile(column) {
+    return this.get(`col${column}File`);
```
stefanpenner: with https://github.com/ember-cli/ember-twiddle/pull/141/files#r37147243 this would likely become:

```js
this.get('column').objectAt(column)
```

Gaurav0:
@stefanpenner Updated. Could you please review again?

app/components/file-editor-column.js
```diff
@@ -18,6 +21,11 @@ export default Ember.Component.extend({
     }
   }),
 
+  lastColumn: computed('col', 'numColumns', function() {
```
stefanpenner: this CP at first glance appears to indicate, it will provide the last column. But after reading the implementation it seems to actually return true/false. Maybe it should be isLastColumn ?

app/gist/controller.js
```diff
@@ -22,14 +29,32 @@ export default Ember.Controller.extend({
...
+   * Column which has the currently focused editor
+   * @type {Number}
+   */
   activeEditorCol: null,
-  col1File: null,
-  col2File: null,
-  col1Active: equal('activeEditorCol','1'),
-  col2Active: equal('activeEditorCol','2'),
+
+  columns: [],
```
stefanpenner: should likely not put a complex object on the prototype, especially if it is meant to hold instance state.

app/gist/controller.js
```diff
@@ -69,20 +94,69 @@ export default Ember.Controller.extend({
...
+  createColumns() {
+    let columns = [];
+    for (let i = 0; i < MAX_COLUMNS; ++i) {
+      let col = (i + 1) + "";
+      columns.pushObject(Column.create({
+        col: col,
+        controller: this
+      }));
+    }
+    this.set('columns', columns);
```
stefanpenner: `createColumns` should likely just be the default implementation of the columns CP getter

app/gist/controller.js
```diff
@@ -69,20 +94,69 @@ export default Ember.Controller.extend({
...
+  },
 
-    if(files.objectAt(0)) {
-      this.setColumnFile(1, files.objectAt(0));
+  /**
+   * Set the initial files in the columns
+   */
+  initializeColumns() {
+    let files = this.get('model.files');
```
stefanpenner: i believe we should likley initialize them on-demand, as part of when the columns CP is computed.

app/components/file-editor-column.js
Gaurav0: Ok, done.

app/gist/controller.js
Gaurav0: Setting in constructor.

app/gist/controller.js
Gaurav0: IDK, computed property getters with side effects make my head hurt.

app/gist/controller.js
Gaurav0: Again, I really don't like setting and initializing things in computed property getters.

Gaurav0:
@stefanpenner Are there any further issues or may I merge this?

PR #145, opened 2015-08-10
Update EmberJS to 1.13.7

blimmer:


rwjblue:
Should also update https://github.com/ember-cli/ember-twiddle/blob/master/blueprints/twiddle.json#L5.

blimmer:
These upgrades are causing all tests to timeout. I don't have time to look at it at the moment so I'll close an open an issue for someone else who might have more time to investigate. Sorry! Was thinking it would be quick.

Issue #35, opened 2015-07-09
Component hbs and js should open simultaneously


joostdevries:

piotrpalek:
@joostdevries I was going to implement this, but I have some questions regarding this.
- should the seperate entries for creating only the component hbs or only the component js really be removed? I mean some people might want to create only a template or js file for the component
- as I understand it after creating the files the dropdown menu (for selecting files) works normally, it shows the seperate js and hbs file for selection (not a single selection which will open both files at once)
- how would you like to handle the file path selection? just open the prompt 2x ?

piotrpalek:
For the 3rd question, I was thinking about letting people only edit part of the path, so if they write `some-component` in the prompt, it would generate `components/some-component.js` and `templates/components/some-component.hbs` files. Less possible to shoot yourself in the foot, but could be too limiting (if twiddle wants to support pod structure simultaneous to the "classic" structure, like cli works).
Could be fine for now though if pod support isn't an issue? Maybe later if pod support is required this can be rethinked and modified.

joostdevries:
@piotrpalek I'd add an option to the file menu called `Create component` which just opens a prompt asking for the component name (instead of path). If people accidentally do enter an extension, just strip it.

I'd leave the separate options in as well.

Gaurav0:
Fixed by #148

PR #148, opened 2015-08-11
add component (both js and hbs) Add File choice

kagemusha:
Takes care of issue #35.

One difference from discussions there is that if user enters file extension, get snarky error msg instead of stripping it off :-), but this can be changed if the latter felt better.

Tests to come.

### DIFF
--- a/app/gist/controller.js
+++ b/app/gist/controller.js
@@ -1,6 +1,7 @@
 import Ember from "ember";
 import config from '../config/environment';
 import Settings from '../models/settings';
+import ErrorMessages from 'ember-twiddle/helpers/error-messages';
 
 const {
   computed: { equal },
@@ -113,7 +114,7 @@ export default Ember.Controller.extend({
     let errorMsg = null;
     if (type.match(/^component/)) {
       if(!path.match(/-[^\/]+$/)) {
-         errorMsg = 'Component file names need to include a hyphen';
+         errorMsg = ErrorMessages.componentsNeedHyphens;
       }
     }
     if (errorMsg) {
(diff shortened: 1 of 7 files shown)


app/gist/controller.js
```diff
@@ -86,6 +86,46 @@ export default Ember.Controller.extend({
...
+  },
+  /*
+   *  Test path is valid.  Presently only tests components are hyphenated.
+   */
+  hasInvalidPath(type, path){
+    let errorMsg = null;
+    if (type === 'component'){
+      let file = path.split('/').slice(-1)[0];
+      if (file.indexOf('.') > -1) {
+        errorMsg = "What part of WITHOUT FILE EXTENSION did you not understand ;-) ?";
```
rwjblue: LOL

rwjblue: Might want to make it *slightly* kinder :smiling_imp:

Gaurav0:
You have a jshint error, and you have some conflicts.

app/gist/controller.js
Gaurav0: @rwjblue @joostdevries Do we want an error message at all or do we want to automatically strip the extension?

kagemusha: @Gaurav0 @rwjblue @joostdevries imagine most of the time people will want to do what stripping accomplishes, so despite my affection for the snarky msg, gonna strip the extension if no objections

Gaurav0: Sounds good.

Gaurav0:
@kagemusha This looks great! Can you do the following:

1. When adding a component, it should open the js file in column 1 and the template in column 2
2. Place the two legacy component options (for individual hbs and js) in a subdropdown (subflyout) menu.

kagemusha:
@Gaurav0 added.  pls take a look.

Gaurav0:
:+1:

Issue #13, opened 2015-05-18
Make it embeddable


joostdevries:
So we can use it in the Ember guides

Gaurav0:
Do we just need to generate code to load it in an iframe, or is there more to this?

stefanpenner:
@Gaurav0 I suspect it may also need some style tweaks.  Typically the embedded examples are presented in a smaller viewport.

joostdevries:
@Gaurav0 yeah when @kevinrudyanderson gets back from his well-earned vacation I hope this can be part of the next design iteration.

PR #154, opened 2015-08-14
Make it embeddable

Gaurav0:
Fixes issue #13

Note: This builds on pull request #141, please review that first.


app/components/file-menu.js
```diff
@@ -23,6 +23,12 @@ export default Ember.Component.extend({
     share() {
       prompt('Ctrl + C ;-)', window.location.href);
     },
+    embed() {
+      let src = window.location.href.split("?")[0];
+      src += "?numColumns=0";
+      let embedCode = `<iframe width="800" height="600" src="${src}"></iframe>`;
```
stefanpenner: the resulting embed values are not correctly sanitized, and could pose a XSS risk.

vector:

1. evil twiddler sends cleverly crafted href, ble to escape the src + frame tag in the interpolated string.

Gaurav0: That's really cool. But we are already using href in Share Twiddle. I'm going to open a PR ASAP and update Ember Twiddle for security reasons.

stefanpenner:
i left comments re: the approach taking by columns [here](https://github.com/ember-cli/ember-twiddle/pull/141)

app/components/file-menu.js
Gaurav0: Actually maybe I'm misunderstanding. According to this, using window.location.href is relatively safe, and generally more safe than alternatives: http://stackoverflow.com/questions/24078332/is-it-secure-to-use-window-location-href-directly-without-validation

Gaurav0:
@stefanpenner @rwjblue @joostdevries Please review.

stefanpenner:
sgtm

Issue #153, opened 2015-08-14
Suggestions relating to identifying twiddles, stopping twiddle escalation


JKGisMe:
So yesterday 3 of us were trying to work out an issue and using twiddles to demonstrate. With all the sharing of twiddle links back and forth it became difficult to tell if you were working on your own twiddle that could be saved, or someone else's twiddle that couldn't be saved. More than a couple times I've seen someone edit a twiddle and think they can share their changes but they didn't realize it was someone else's twiddle and lost what they did. Also when multiple people are working on twiddles everyone ends up forking twiddles so the number of twiddles tends to increase exponentially just to make small edits.

Some solutions to this that came to mind were: 
- URL slugs with the title of the twiddle so when there are 3-4 twiddle links in Slack you can tell which one is which. 
- Some kind of indication on the top bar (where the title and your user name are displayed) that shows if it's your twiddle or someone elses.
- an indication of whether the twiddle is save-able/forkable.
- setting a twiddle as locked so only the owner can edit & save or as public so anyone can edit & save

We ended up linking a twiddle to an issue in the ember repo. So I don't want to change the twiddle accidentally myself and no longer have it demonstrate the issue when people view the ember issue. I think being able to set a 'read only' mode where you have to maybe uncheck a checkbox to edit the twiddle(or fork it when forking ones own twiddle is available) would be nice in that kind of situation.

Gaurav0:
So on canary.ember-twiddle.com you keep and save your changes when you fork. Does this work for you?

JKGisMe:
This definitely helps. Perhaps when you are not the owner of the current twiddle the 'save' menu item should be disabled. There is the indicator that says 'saved' or 'unsaved' but it's easy to overlook if you've mistakenly tried to save on someone else's twiddle.

PR #157, opened 2015-08-16
Improve Forking and Saving

Gaurav0:
1. The fork menu option now no longer appears unless you do not own the gist
2. The save option forks if you do not own the gist

Relevant to issue #153

### DIFF
--- a/app/components/file-menu.js
+++ b/app/components/file-menu.js
@@ -1,8 +1,14 @@
 import Ember from "ember";
 
+const { computed } = Ember;
+
 export default Ember.Component.extend({
   tagName: 'li',
   classNames: ['dropdown'],
+
+  belongsToUser: computed('model.ownerLogin', 'session.currentUser.login', function() {
+    return this.get('model.ownerLogin') === this.get('session.currentUser.login');
+  }),
 
   actions: {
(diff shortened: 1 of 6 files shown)


stefanpenner:
SGTM – :shipit:

chancancode:
Can we add back a way to fork your own twiddle? I find I wanted to do it quite often for various reasons: creating stable "steps" I can embed in a long tutorial, having a base template to build off of, etc. I would even take something like "hold down option key to change save to fork" :) (But I think it is still quite legitimate that you would want to fork your own twiddle, so I don't think it needs to be a hidden/advanced feature)

Also, I have a UI suggestion along the same lines: The file menu currently mixes commands that operates on file-level (add/move/delete) vs things that operates on the twiddle level (new/save/fork/share). Maybe move the latter above the divider to separate the two sets of things?

Gaurav0:
The github api does not provide a way to fork one's own gist, it just returns an error if you try to do so. An issue is open for duplicating one's own gist, but it will likely be a copy, not a fork with a link to the original or ability to PR.

chancancode:
:+1: that actually works well for my purpose, thanks!

PR #159, opened 2015-08-18
better title input

JKGisMe:
press enter to focus out
clear default text on focus in
replace default text on focus out

### DIFF
--- a/app/components/title-input.js
+++ b/app/components/title-input.js
@@ -6,23 +6,25 @@ export default Ember.Component.extend({
 
   //proxy clicks to input focus
   click() {
-    this.$('input').focus();
-  },
-
-  enter() {
-    this.$('input').blur();
+    this.$('input').focusin();
   },
 
   actions: {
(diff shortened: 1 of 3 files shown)


app/components/title-input.js
```diff
@@ -9,13 +9,30 @@ export default Ember.Component.extend({
...
 
+  enter() {
+    this.$('input').blur();
+  },
+
   actions: {
     inputFocusIn() {
+      if(this.$('input').val()){
+        this.$('input').val('');
+    }
```
Gaurav0: 1. Please fix indentation.
2. Won't this delete a title that isn't the default text?

JKGisMe:
I intended it to clear everything because regardless of what's in there it's annoying to have to backspace before you can type but on 2nd thought I just selected the existing text so it could be easily overwritten but not destroyed.

Gaurav0:
Works for me! :+1:

PR #160, opened 2015-08-18
stops the corners of user avatar from being cut off

JKGisMe:
Before:
image

After:
image

This probably looks really different with different avatars. My ears always get cut off at the tips.


Gaurav0:
:+1:

PR #161, opened 2015-08-18
File tree component

Gaurav0:
This PR does the following:
1. Adds a file-tree component
2. Hides it by default
3. Gives it the functionality of opening files in the active file editor panel
4. Creates a more complete full screen mode, rather than just numColumns=0
5. Hides the file-tree in that full screen mode
6. Adds a way to exit full screen mode


Gaurav0:
@joostdevries @rwjblue @stefanpenner  This PR is now complete. Please review.

PR #169, opened 2015-08-22
set first column active on load

kagemusha:
Small change, but currently when page loads and code panes are greyed out, can give impression that twiddle is disabled.  This seeks to highlights that is editable and people can jump right in.

### DIFF
--- a/app/gist/controller.js
+++ b/app/gist/controller.js
@@ -24,6 +24,7 @@ export default Ember.Controller.extend({
     this._super(...arguments);
     this.createColumns();
     this.setupWindowUpdate();
+    this.set('activeEditorCol', '1');
   },
 
   /**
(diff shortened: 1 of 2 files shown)


app/gist/controller.js
```diff
@@ -113,6 +113,7 @@ export default Ember.Controller.extend({
         controller: this
       }));
     }
+    columns.objectAt(0).set('active', true);
```
Gaurav0: Column#active is a computed property with no setter.

The correct way to make a column active is to set activeEditorCol on the controller.

Gaurav0:
This PR causes the first column to always be active.

kagemusha:
@Gaurav0  thanks for the insight.  edited based on your suggestion.

Issue #172, opened 2015-08-25
Make blueprint filepaths consistent, either pods or traditional


Gaurav0:

Gaurav0:
Fixed by #175

PR #175, opened 2015-08-26
Use pod format consistently

Gaurav0:
Fixes issue #172


rwjblue:
Looks good, can we have a test for the logic added in `addComponent`?

Gaurav0:
@rwjblue Test added.

PR #186, opened 2015-09-03
Update to new serializer API

Gaurav0:
As to why this new API is any better, I don't know. Just seems more confusing to me.


app/gist/route.js
```diff
@@ -15,6 +15,8 @@ export default Ember.Route.extend({
     saveGist (gist) {
       var newGist = gist.get('isNew');
       if (!newGist && gist.get('ownerLogin') !== this.get('session.currentUser.login')) {
+        console.log(gist.get('ownerLogin'));
```
joostdevries: beep

stefanpenner: boop

PR #190, opened 2015-09-07
better sign in that fixes #183 & hack that fixes #187 ....again

### LINKED ISSUE #183: Change Anonymous menu drop down -> "Sign In"
Having the Anonymous drop down with a "Sign In" option makes it less clear that there is "sign in" option available.  Could we change it to say "Sign In" by default and then change it to a drop down menu once someone is logged in?

### LINKED ISSUE #187: Main Header is Cut Off
I remember pointing out when we worked on the redesign that this was a bug, so it's probably still there from before. Size of the browser window doesn't seem to matter.

JKGisMe:
### Current
twiddle-signin-old

### This PR
twiddle-signin-new

I moved the github icon to the right because it looked disjointed on the left somehow. Once signed in, the user's avatar is unchanged.


app/components/user-menu.js
```diff
@@ -4,9 +4,7 @@ export default Ember.Component.extend({
   tagName: 'ul',
   classNames: ['nav', 'nav-pills', 'user-menu'],
 
-  userName: Ember.computed('session.isAuthenticated', function() {
-    return this.get('session.isAuthenticated') ? this.get('session.currentUser.login') : "Anonymous";
-  }),
+  userName: Ember.computed.alias('session.currentUser.login'),
```
stefanpenner: computed.readOnly

stefanpenner:
looks great

JKGisMe:
Haha as soon as I submitted the PR I realized that would have been better. I'll change it.

app/gist/new/route.js
```diff
@@ -7,7 +7,7 @@ export default GistRoute.extend({
   model () {
     this.store.unloadAll('gistFile');
 
-    var model = this.store.createRecord('gist', {description: 'New Twiddle'});
+    var model = this.store.createRecord('gist', {description: 'New Twiddle '});
```
Gaurav0: Adding a space here won't fix it if the user changes the name.

app/templates/components/user-menu.hbs
```diff
@@ -1,23 +1,25 @@
...
+    <li class="dropdown">
+      <a class="dropdown-toggle" data-toggle="dropdown" href="#">
+        {{userName}} <b class="caret"></b>
+      </a>
+      <ul class="dropdown-menu dropdown-menu-right">
+        <li><a {{action 'signOut'}} class="test-sign-out">Sign out</a></li>
+      </ul>
+    </li>
+  {{else}}
+    <li><a {{action 'signInViaGithub'}} class="test-sign-in sign-in">Sign in with Github</a></li>
```
Gaurav0: Can you make this just Sign in? It takes up too much space on mobile and small screens.

app/gist/new/route.js
JKGisMe: Hm, I wasn't aware the problem happened when the title had been changed. Even if the named has been changed and then later reverted back to `New Twiddle`.  Please show steps to reproduce a once edited title getting cut off.

Gaurav0: So you have to

1. Change the title
2. Save the gist
3. Load Ember Twiddle with the saved gist.

JKGisMe: I tried that before and just tried it again and the title was not cut off. I tried it both logged in and logged out. This issue is unpredictable enough it's almost like a practical joke!

Gaurav0: Yes, I know. Sorry. Maybe instead of adding a space to the value you can add a space to a computed property based on value and show that?

PR #198, opened 2015-09-22
Fix tree does not nest than one level

### LINKED ISSUE #197: File tree does not nest
Create a nested file e.g. application/error/template.hbs

File tree shows folder named application/error and template.hbs inside it rather than nested tree.

Gaurav0:
Fixes #197


rwjblue:
:+1:

PR #202, opened 2015-09-24
Fix New Twiddle doesn't change pane content

### LINKED ISSUE #201: New Twiddle doesnt change pane content
1. Open one of your twiddles (didn't try it with one that wasn't mine)
2. Click File -> New Twiddle
3. Only the output pane changes to reflect the new twiddle while the left and middle panes keep showing the previous twiddle content.

Gaurav0:
Fixes #201

### DIFF
--- a/app/gist/controller.js
+++ b/app/gist/controller.js
@@ -124,6 +124,16 @@ export default Ember.Controller.extend({
   },
 
   /**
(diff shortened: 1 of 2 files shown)


rwjblue:
A test would be nice, but this looks good.

Issue #78, opened 2015-07-24
Duplicate twiddle you already own


Gaurav0:

pangratz:
This one can be closed since #215 has been merged ...

PR #215, opened 2015-10-12
Add possibility to copy own twiddle

pangratz:
If a gist owned by the authenticated user is shown, a new menu item
"Copy Twiddle" allows to copy the current files into a new Twiddle. Note
that the newly created Twiddle is not yet saved.

By this, Twiddles (e.g. Ember setup for a specific version) can be used
as a starting point without having to re-setup the whole Twiddle again
and again.

---

This addresses #78 and https://github.com/ember-cli/ember-twiddle/pull/157#issuecomment-138380066.

### DIFF
--- a/tests/acceptance/gist-test.js
+++ b/tests/acceptance/gist-test.js
@@ -3,6 +3,7 @@ import { module, test } from 'qunit';
 import startApp from 'ember-twiddle/tests/helpers/start-app';
 import { findMapText } from 'ember-twiddle/tests/helpers/util';
 import ErrorMessages from 'ember-twiddle/helpers/error-messages';
+import { stubValidSession } from 'ember-twiddle/tests/helpers/torii';
 
 
 const firstColumn = '.code:first-of-type';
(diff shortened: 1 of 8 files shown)


rwjblue:
:+1: - LGTM

pangratz:
Aaaand it's live on http://canary.ember-twiddle.com. Oh boy, this workflow is awesome! :dancers:

tests/acceptance/gist-test.js
```diff
@@ -223,3 +224,37 @@ test('unsaved indicator', function(assert) {
...
+  });
+
+  click('.test-copy-action');
+
+  andThen(function() {
+    assert.equal(find('.title input').val(), "New Twiddle", "Description is reset");
+    assert.equal(find('.test-unsaved-indicator').length, 1, "Unsaved indicator appears when gist is copied");
+    assert.equal(find('.test-copy-action').length, 0, "Menu item to copy gist is not shown anymore");
+  });
+});
```
Gaurav0: Really cool test!

pangratz: Thx! But it currently misses assertions that all files have been copied correctly. Do you have any suggestions? Would it be sufficient to run the twiddle and check if the output is the same as the original one?

Issue #216, opened 2015-10-12
Add Copy Route


rwjblue:
As of https://github.com/ember-cli/ember-twiddle/pull/215 we can now copy a twiddle very easily. I would like a way to bookmark a given twiddle so that it immediately launches into a freshly copied twiddle.

Today, I keep a JSBin for every Ember version (for issue reproduction and quick testing purposes) and launch that base JSBin to get a quick clone.  It would be awesome if I could launch `ember-twiddle.com/<some-gist-id>/copy` where the behavior would roughly be:

```javascript
// app/gist/copy/route.js

import Ember from "ember";
import GistRoute from "ember-twiddle/routes/gist-base-route";

export default GistRoute.extend({
  model (params) {
    this.store.unloadAll('gistFile');

    return this.store.find('gist', params.id);
  },

  afterModel() {
    this.transitionTo('gist.new', {
      queryParams: {
        copyCurrentTwiddle: true
      }
    });
  }
}
```

At that point it would be trivial to store bookmarks for each twiddle that link directly to a `copy`...

---

@pangratz - I'm unsure how much work it would be beyond the snippet above. Since this builds off what you did in #215, would you wanna take a crack at it?

pangratz:
What do you think of `ember-twiddle.com/<some-gist-id>?copy=true` ?

rwjblue:
HA!, I actually had typed that first then deleted it ;P

rwjblue:
Fixed by #217

PR #217, opened 2015-10-13
Copy gist route

pangratz:
This PR creates a new route `/:gist/copy` which is a shortcut to copy a gist into a new Twiddle.

This addresses #216.


Gaurav0:
:+1:

rwjblue:
:heart:

Issue #188, opened 2015-09-06
Add "FEATURES" section to twiddle.json


chancancode:
so people can use experimental features on their fiddles

Here is a quick hack to do this today: https://gist.github.com/chancancode/225639b3144dd82f3b6d

The key is to add an `EmberENV` override as a dependency before Ember: https://cdn.rawgit.com/chancancode/5b7b95b6268d51cdcaae/raw/92e0654083305c73cd601381a6cea705be9f8890/hax.js

pangratz:
Since #219 got merged, I guess this issue can be closed :tada:

rwjblue:
Thanks @pangratz!

PR #219, opened 2015-10-15
Add possibility to specify EmberENV in twiddle.json

pangratz:
The value of the top-level `EmberENV` key of twidde.json is set as
`window.EmberENV` in the Twiddles' application index.html. By this,
Twiddle can be created which make use of features in Ember, which are
put behind a feature flag:

    // twiddle.json
    {
      "EmberENV": {
        "FEATURES": {
          "ember-htmlbars-component-generation": true
        }
      }
    }

---

This addresses #188


Gaurav0:
:+1:

Issue #220, opened 2015-10-16
"Run Now" should reload the application even if no changes to the files have been made


Gaurav0:

rwjblue:
Agreed. This has bitten me more than once.

Gaurav0:
Fixed by #221

PR #221, opened 2015-10-18
"Run now" reloads app

pangratz:
The used `{{dummy-demo-app}}` component only reloads the application if
the passed `html` property changes. If nothing changes in the underlying
app, succeeding clicks on "run now" will do nothing, as the HTML doesn't
change.

To force enable the reload of the application, this commit adds a new
`<meta name="build-at" >` tag, which' value is set to the milliseconds
since the Unix Epoch. By this, the HTML changes everytime the app is
build and now the "run now" button always reloads the application.

---

This addresses #220

### DIFF
--- a/tests/acceptance/run-now-test.js
+++ b/tests/acceptance/run-now-test.js
@@ -1,0 +1,48 @@
+import Ember from 'ember';
+import { module, test } from 'qunit';
+import startApp from 'ember-twiddle/tests/helpers/start-app';
+
+module('Acceptance | run now', {
+  beforeEach: function() {
+    this.application = startApp();
+  },
+
+  afterEach: function() {
+    Ember.run(this.application, 'destroy');
+  }
+});
+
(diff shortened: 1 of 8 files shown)


tests/acceptance/run-now-test.js
```diff
@@ -0,0 +1,48 @@
...
+    outputPane().find('input').val('new value');
+
+    assert.equal(outputPane().find('input').val(), 'new value');
+  });
+
+  andThen(function() {
+    find(".run-now").click();
+  });
+
+  waitForLoadedIFrame();
```
Gaurav0: I think you may need to move this inside previous andThen

pangratz: You are right, nice catch.

Gaurav0:
:+1:

rwjblue:
Awesome!

PR #222, opened 2015-10-19
Add border-radius to profile pic of logged in user

pangratz:
This makes it consistent with the profile pic rendered on GitHub.

Subtle change but makes the appearance more smoother IMHO.

---

Before

_before

After

_after


Gaurav0:
@JKGisMe Could you check in Canary if this cuts off your ears again?

PR #223, opened 2015-10-21
[UI] highlight current row in list of twiddles

pangratz:
highlight-twiddle-row mov

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -471,4 +471,8 @@ body {
 .saved-twiddles {
   width: 80%;
   margin: 10px auto;
+
+  tr:hover {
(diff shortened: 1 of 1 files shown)


rwjblue:
LGTM

@Gaurav0 - :+1: / :-1: ?

alexspeller:
Is there a reason this isn't just `tr:hover { }` ?

pangratz:
> Is there a reason this isn't just tr:hover { background-color: $ember-orange; } ?

The only reason I can think of is that I haven't thought of that. It's crazy what you can do with plain CSS :stuck_out_tongue_winking_eye:

Thanks @alexspeller, will update to this much nicer solution!

PR #230, opened 2015-10-23
Add empty FEATURES to twiddle.json

Gaurav0:
This makes it obvious how to add feature flags to a new twiddle at the cost of some extra bytes per twiddle.


rwjblue:
Looks good.  I'd like to also (not right away) add the ability to strip comments from the JSON file so that we can leave some inline comments indicating what each section does.

PR #233, opened 2015-10-24
Implement logout on /twiddles route

pangratz:
A logout on the /twiddles route navigates to the page where a new, empty
twiddle is shown.


Gaurav0:
:+1:

PR #238, opened 2015-10-26
Use arrow function to preserve this.

rwjblue:
This previously caused an error when saving because `this` wasn't being preserved.

Prior error:

screenshot


Gaurav0:
:+1:

Issue #234, opened 2015-10-25
Fail loudly when saving gist does not work.


workmanw:
So my twiddle ended up in a state where it said it was saving my Gist to Github and it actually wasn't. 

Here is what my console looks like:
![image](https://cloud.githubusercontent.com/assets/825487/10713633/df4d0200-7a97-11e5-8e2e-4d017c1f8dcb.png)

Here is the network request:
![image](https://cloud.githubusercontent.com/assets/825487/10713640/1e7c3432-7a98-11e5-86a7-bd522e3821e5.png)

Here is the PUT data:
![image](https://cloud.githubusercontent.com/assets/825487/10713649/7b89a3a8-7a98-11e5-890d-9e2755bd7b70.png)

I'm very lucky I sent this twiddle to a friend for review first. Otherwise I would have lost two hours of work. I actually don't see the duplicate file ... I don't know how to help you reproduce this. But you should really make this a priority to alert users when a Save to Github fails!!!!

workmanw:
Well upon further inspection I found the duplicate file. It seems to be internal data corruption of some kind.

![image](https://cloud.githubusercontent.com/assets/825487/10713659/1f8a95f2-7a99-11e5-9792-7f98efb6e37d.png)

Please fix this.

workmanw:
Alright so I'm sitting here doing my best to help debug this since I've got this active Ember Twiddle app running. I think there is some flaw in the way files get deleted by Ember-Twiddle. 

I'm using a new twiddle in an incognito window and It seems that deleting a file doesn't save properly to the gist. I was deleting and moving files so that I did not use the pod structure. I suspect this is related to the cause of this issue.

Gaurav0:
I am so sorry for your issue.

I tried the following:

1. Opened an incognito window
2. Signed in
3. Deleted application/controller.js
4. Created a new application/controller.js
5. Saved

I was successful.

If you can find a repeatable set of actions that can get you into this state, please let me know.

workmanw:
I am not able to reproduce this exact issue. Believe me though, it happened and I tried and tried.

I did reproduce something that might be related.

1. Opened new twiddle and sign in.
2. Save
3. Move "application/controller.js" to "controllers/application.js"
4. Save
5. Delete "controllers/application.js"
6. Save

Now visit the Gist link and notice that "controllers.application.js" is still in the Gist. But the Twiddle thinks it's been deleted.

rwjblue:
[...]

PR #239, opened 2015-10-26
Loud save error

Gaurav0:
Attempt to always show a loud error if saving or forking fails.

Issue #234


rwjblue:
Looks good to me.  I think we need to add some acceptance tests around saving (which I couldn't find on a quick review), but that can be an issue for another day...

PR #241, opened 2015-10-26
Fix failing tests

pangratz:
Executing the tests locally threw an error due to run-loop issues with `ember-resize`. This has been fixed upstream in https://github.com/tim-evans/ember-autoresize/pull/26.

---

Parsing an empty string for `twiddle.json` throws an error at the end of the `visting /twiddles` acceptance test, which somehow isn't failing the test but the error is logged in the console of Web Inspector.


Gaurav0:
:+1:

PR #243, opened 2015-10-27
Added wrapper around iframe to make it responsive.

### LINKED ISSUE #242: make embed responsive
The fixed size of the iframe is unfortunate. If there is more room available it does not utilise it. If less, the iframe is cut off.

brianally:
Fixes #242. The embed markup includes a wrapper div and some minimal inline styles on both it and the iframe to allow it to use the full width of the containing element. I chose 16:9 ratio. I tried 4:3 but found it to be too tall. This could be configurable, although it would mean an extra step for the user in creating the embed code.

### DIFF
--- a/app/components/file-menu.js
+++ b/app/components/file-menu.js
@@ -33,13 +33,15 @@ export default Ember.Component.extend({
       prompt('Ctrl + C ;-)', window.location.href);
     },
     embed() {
-      let src = window.location.href.split("?")[0];
-      src += "?fullScreen=true";
-      let iframe = document.createElement("iframe");
+      let src = window.location.href.split('?')[0];
+      src += '?fullScreen=true';
+      let responsive = document.createElement('div');
+      responsive.style.cssText = 'position:relative;height:0;overflow:hidden;max-width:100%;padding-bottom:56.25%;'; // 16:9
+      let iframe = document.createElement('iframe');
       iframe.src = src;
-      iframe.width = 800;
-      iframe.height = 600;
-      let embedCode = iframe.outerHTML;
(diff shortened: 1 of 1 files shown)


brianally:
I just noticed i needlessly changed a couple of lines. I saw later that the rest of the file uses single quotes so changed my doubles, changing a couple of originals as well.

app/components/file-menu.js
```diff
@@ -33,13 +33,15 @@ export default Ember.Component.extend({
       prompt('Ctrl + C ;-)', window.location.href);
     },
     embed() {
-      let src = window.location.href.split("?")[0];
-      src += "?fullScreen=true";
-      let iframe = document.createElement("iframe");
+      let src = window.location.href.split('?')[0];
+      src += '?fullScreen=true';
+      let responsive = document.createElement('div');
+      responsive.style.cssText = 'position:relative;height:0;overflow:hidden;max-width:100%;padding-bottom:56.25%;'; // 16:9
```
Gaurav0: Many people who are embedding this in a page will want to specify the width and the height. How would they do this?

brianally: By setting the width of the containing element, ie. what they're embedding it into. With this change, the iframe takes up the full width of the "column" it's in and the height adjusts proportionally. [See here](https://css-tricks.com/NetMag/FluidWidthVideo/Article-FluidWidthVideo.php) for more info on how this works.

As it stands now, users cannot set the dimensions of the iframe unless they edit the markup before pasting into their own document.

Gaurav0:
:+1:

Issue #118, opened 2015-08-01
Ability to type route name in the twiddle address bar


abidmix:
The default url bar at present is a div which is ready only. It would be nice to type a known route and be taken to it during a demo in twiddle. I have created a gist and a  twiddle http://ember-twiddle.com/ec81fea57d0555c1859a  to demonstrate the desired functionality

JKGisMe:
+1 
I went to go experiment with different path related stuffs in a twiddle and realized I couldn't really interact with it much to see if what I did in the router was working how I wanted. Making the existing url bar interactive would be super cool.

abidmix:
I downloaded the source code, rummaged the internals and was able to achieve the functionality, with some help from my EmberNYC meetup colleagues. I will be submitting a pull request soon. Check out the interactive url bar in action at  http://ember-twiddle.com/ec81fea57d0555c1859a

rwjblue:
Awesome!

Gaurav0:
@abidmix So to turn this into a PR, we need to make the existing url bar a textfield input box, and when the user types, post a message to the app inside the iframe, and in the iframe, on postmessage, transition the router to the route entered.

abidmix:
Thanks @rwjblue.   @Gaurav0 ,I agree. I will be setting up my dev environment in accordance wth the instructions in the contributing.md document. I will shout if I need help in wiring the postmessage to the iframe.

JKGisMe:
I love the error message `Oops.... it seems you forgot to define this route in your router.js`

alexspeller:
I suspect the simplest / cleanest way to implement this might be by injecting a custom location implementation

Gaurav0:
Fixed by #248

PR #248, opened 2015-10-27
Two-way address bar

pangratz:
This change makes the address bar for the demo app functional, so
entering a new URL and pressing enter triggers a transition to that URL
in the demo app.

This addresses #118.

---

The basic idea is to inform the demo app via `postMessage` about a newly entered URL and do a `router.transition(newUrl)`. The current state looks as follows:

my-name-is-url mov

---

The tests pass locally when accessed via `localhost:4200/tests` but they **do not in PhantomJS**. I don't know enough phantom to solve this, but this might be related to security issues?


Gaurav0:
Closing in favor of #257

Issue #245, opened 2015-10-27
missing gh logo


brianally:
![screen shot 2015-10-27 at 3 13 38 am](https://cloud.githubusercontent.com/assets/8657255/10751779/be104d72-7c59-11e5-83a4-05fc03209ca6.png)

Gaurav0:
Mixed Content warning appears in console.

brianally:
Strange. I just saw the mixed content warning as well. But I didn't last night. That was the first thing I looked for; figured the image was just missing when I didn't see anything in the console. I reloaded just now and the warning is not there, but the image is still not appearing.

brianally:
duh! The mixed content warning is only going to appear when using SSL.

The svg shows up now in my local fork (it wasn't earlier) but still not live. The toolbar bg path begins with "../images/" whereas the GH svg has "images/".

brianally:
It doesn't appear that SRI is what is prepending the assets URI for the toolbar bg. But it is being prepended, so the simplest fix looks like removing the GH logo inline image and setting a bg in CSS as the toolbar has. I can do that now and check it works.

Gaurav0:
Please verify this works in canary.ember-twiddle.com

brianally:
Yes and no. I can see the logo just fine but the styles look like:

    .toolbar .user-menu .user-avatar.unauthenticated {
      background-image: url("../images/github32-inverse-faf2ee.svg");
    }

    .toolbar {
      background: #e1563f url("//canary-assets.ember-twiddle.com/images/navigation_background-c1437f14bdf59ef96874cb26a83031fe.png") repeat scroll center top;
      border-bottom: 1px solid #aa412f;
      box-shadow: 0 0 3px rgba(0, 0, 0, 0.3);
      font-size: 13px;
      height: 60px;
    }

Is there a regex somewhere that needs "svg" added to get itself prepended and fingerprinted?

Gaurav0:
I googled it. See the 'extensions' option under 'fingerprint'. http://www.ember-cli.com/asset-compilation/#fingerprinting-and-cdn-urls

brianally:
Ack! I was looking for where the "js", "css", and "png" were configured. Defaults ftw.

Gaurav0:
Fixed by #261

PR #249, opened 2015-10-27
Changed GH logo to be a bg image in css rather than inline image

brianally:
… to ensure that the URL is correct.


brianally:
Fixes #245

Gaurav0:
:+1:

PR #257, opened 2015-10-28
Two way address bar

Gaurav0:
This updates PR #248 with tests that pass reliably in PhantomJS

@joostdevries @rwjblue @alexspeller @stefanpenner Please review.


app/components/address-bar.js
```diff
@@ -0,0 +1,9 @@
+import Ember from "ember";
+
+export default Ember.Component.extend({
+  classNames: ["url-bar"],
+
+  change(event) {
+    this.sendAction("urlChanged", event.target.value);
+  }
```
pangratz: Just to be sure, this will be fired when a new URL is entered and the element is blurred, right? In the previous commit, the action is only fired, when enter is pressed...

pangratz:
Apart from my comment, this looks good! Thanks for fixing my failing tests @Gaurav0 :smile: 

I guess the waiting for the iframe to be loaded should eventually be refactored using [`Ember.Test.registerWaiter`](http://emberjs.com/api/classes/Ember.Test.html#method_registerWaiter), but this is out of the scope for this PR.

app/components/address-bar.js
Gaurav0: I'm not sure. But it seems that in testing, fillIn never fires the enter event, whatever you do.

http://stackoverflow.com/questions/28586314/trigger-an-inputs-action-from-an-acceptance-test#comment45530171_28596428

tests/acceptance/routing-test.js
```diff
@@ -12,51 +12,147 @@ module('Acceptance | routing', {
...
+
+    return new Ember.RSVP.Promise(function (resolve) {
+      Ember.run.next(function() {
+        assert.equal(find(addressBar).val(), '/', "Correct URL is shown in address bar 0");
 
-    iframe_window.click(iframe_window.find(aboutLink));
+        iframe_window = outputPane();
+        iframe_window.click(iframe_window.find(aboutLink)).then(resolve);
+
+        done();
```
stefanpenner: `resolve();

tests/acceptance/routing-test.js
```diff
@@ -12,51 +12,147 @@ module('Acceptance | routing', {
...
+    return new Ember.RSVP.Promise(function(resolve) {
+      Ember.run.next(function() {
+        assert.equal(outputContents(outletText), 'About Page', 'About Link leads to About Page being displayed');
+        assert.equal(find(addressBar).val(), '/about', "Correct URL is shown in address bar 1");
 
-    iframe_window.click(iframe_window.find(indexLink));
+        iframe_window.click(iframe_window.find(indexLink)).then(resolve);
+
+        done();
+      });
```
stefanpenner: ^

tests/acceptance/routing-test.js
```diff
@@ -12,51 +12,147 @@ module('Acceptance | routing', {
...
+
+    return new Ember.RSVP.Promise(function(resolve) {
+      Ember.run.next(function() {
+        assert.equal(find(addressBar).val(), '/', "Correct URL is shown in address bar 0");
+
+        click(addressBar);
+        fillIn(addressBar, '/about');
+        keyEvent(addressBar, 'keyup', 13).then(resolve);
+
+        done();
```
stefanpenner: ^

tests/acceptance/routing-test.js
```diff
@@ -12,51 +12,147 @@ module('Acceptance | routing', {
...
+    return new Ember.RSVP.Promise(function(resolve) {
+      Ember.run.next(function() {
+        assert.equal(outputContents(outletText), 'About Page', 'Changing the URL to /about and pressing enter leads to the About Page being displayed');
+        assert.equal(find(addressBar).val(), '/about', "Correct URL is shown in address bar 1");
+
+        click(addressBar);
+        fillIn(addressBar, '/');
+        keyEvent(addressBar, 'keyup', 13).then(resolve);
+
+        done();
```
stefanpenner: ^

tests/acceptance/routing-test.js
```diff
@@ -12,51 +12,147 @@ module('Acceptance | routing', {
...
+  });
+
+  andThen(function() {
+    const done = assert.async();
+
+    Ember.run.next(function() {
+      assert.equal(outputContents(outletText), 'Main Page', 'Changing the URL to / and pressing enter leads to the Main Page being displayed');
+      assert.equal(find(addressBar).val(), '/', "Correct URL is shown in address bar 2");
+
+      done();
```
stefanpenner: ^

tests/acceptance/routing-test.js
```diff
@@ -12,51 +12,147 @@ module('Acceptance | routing', {
...
 
-  runGist(files);
+  runGist(TWIDDLE_WITH_ROUTES);
 
   andThen(function() {
-    iframe_window = outputPane();
+    const done = assert.async();
+
+    return new Ember.RSVP.Promise(function (resolve) {
+      Ember.run.next(function() {
```
stefanpenner: rather then the `next` here, which may be racy, we could make the async communication with the remote frame part of the test-waiters. This would make `andThen` "just work"

Gaurav0: Please explain further.

app/components/address-bar.js
Gaurav0: @pangratz Yes, blurring now makes the page switch.

tests/acceptance/routing-test.js
stefanpenner: http://emberjs.com/api/classes/Ember.Test.html#method_registerWaiter

usage in liquid-fire (for an example)  https://github.com/ef4/liquid-fire/blob/8c076f0f5edb2b44d2e16437a84a290c946a0748/addon/transition-map.js#L17-L19

Gaurav0:
@joostdevries @rwjblue @alexspeller @stefanpenner @pangratz All comments addressed. Please let me know if ok to merge.

joostdevries:
lgtm :+1:

pangratz:
wohoo, party! :tada:

tests/acceptance/routing-test.js
```diff
@@ -5,58 +5,136 @@ import startApp from 'ember-twiddle/tests/helpers/start-app';
...
+      this._waiter = () => {
+        return window.messagesWaiting === 0;
+      };
+      Ember.Test.registerWaiter(this._waiter);
+    };
   },
 
   afterEach: function() {
+    window.messagesWaiting = 0;
+    if (this._waiter) {
```
stefanpenner: The idea was for the app code to do the waiting code, as it understands the async more closely.

Gaurav0: The problem with that is that one app sends the communication and a different app receives it.

pangratz: Maybe an event on `willTransition` should be send in the demo app as well. I will try to look into this tomorrow...

joostdevries:
@stefanpenner @Gaurav0 I think this is a huge step forward. A next iteration on this would ideally restart the app with the provided route as initial state (so you can more effectively debug model hook issues). cc @rwjblue

PR #266, opened 2015-10-30
Disable Channels from Dependencies Menu

Gaurav0:
This disables the channels options in the dependencies menu, because they don't work over https.


Gaurav0:
Closing in favor of #267

PR #267, opened 2015-10-30
Use caching proxy server ember.alexspeller.com for channels

Gaurav0:
Fixes mixed use warning when using channels


alexspeller:
:+1:

joostdevries:
Maybe a cleaner solution is to setup a subdomain on ember-twiddle.com?

joostdevries:
:+1: for now. will look for better solution later

PR #270, opened 2015-11-04
Fix initializer warning when testing against Ember 2.1.

rwjblue:
Still one more to fix (container-debug-adapter).


Gaurav0:
:+1:

PR #275, opened 2015-11-11
Allow user to specify initial route via query parameter

Gaurav0:
Also improves api of runGist to allow for specifying gist parameters other than files.


alexspeller:
:+1: looks good to me

PR #277, opened 2015-11-12
Include icons in user menu

knownasilya:
This allows clicking the icons (after login, and before) to
access the actions that they provide.

So the github icon is now clickable for signin and the user's avatar for
opening the user menu.

screen shot 2015-11-12 at 11 31 55 am
screen shot 2015-11-12 at 11 31 45 am


Gaurav0:
:+1:

PR #282, opened 2015-11-16
Update ember-cli-jstree to fix 'calling set on destroyed object' errors in Chrome

Gaurav0:
Note: Some of the jstree related tests still sporadically fail in Chrome. @ritesh83 can you please look into this?


ritesh83:
@Gaurav0 I checked out the 'set_destroyed' branch and ran the tests twice. The tree tests seem to be passing every time. Not sure why they would fail sporadically.

PR #289, opened 2015-12-02
Escape backticks in templates.

rwjblue:
Prior to this, inclusion of a backtick in a template would throw an
error.


Gaurav0:
:+1:

Issue #123, opened 2015-08-01
Share with specific files shown


HeroicEric:
I usually use emberjs.jsbin.com to quickly show little examples to people or try them out for myself. When sharing with somebody else, it's important to be able to link them to the code that is specific to what I'm trying to show.

When trying to share twiddles with others, I feel like I need to also tell them which files to open up to see what I'm talking about.

For example, I want to link somebody to http://ember-twiddle.com/da9d9673a1102ba0a476 but I'd like to have the `index/route.js` and `index/template.hbs` be the files that are opened when they get there.

I don't think this would be too difficult to implement right now using queryParams but it may get more complicated if https://github.com/ember-cli/ember-twiddle/issues/122 is done

acorncom:
This probably falls under the same general issue.  It'd be nice to be able to highlight certain lines of code within a Twiddle (like can be done with Github). Could get complicated if we tried to highlight multiple lines in multiple panes though ...

HeroicEric:
@acorncom I think that would be a separate feature to be worked on separately. That would add a lot to the complexity of my original issue

acorncom:
@HeroicEric Sounds good.  Opening a separate issue, sorry to pollute this one ...

HeroicEric:
This was completed by https://github.com/ember-cli/ember-twiddle/pull/296

PR #296, opened 2015-12-17
Specify open files via query paramter

Gaurav0:
This is for issue #123.

### DIFF
--- a/app/gist/controller.js
+++ b/app/gist/controller.js
@@ -16,9 +16,10 @@ export default Ember.Controller.extend({
   dependencyResolver: inject.service(),
   notify: inject.service('notify'),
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
Should link-to that changes the query param open then files in question? Or is that out of scope

Gaurav0:
@stefanpenner I think that should already work. Let me check.

Gaurav0:
@stefanpenner A link-to within the outer application does indeed open the files. You can't do it from the twiddle itself, though. And I think that is ok.

stefanpenner:
> @stefanpenner A link-to within the outer application does indeed open the files. You can't do it from the twiddle itself, though. And I think that is ok.

:+1: sounds great.

alexspeller:
:tada:

Issue #290, opened 2015-12-03
navigation in tree appears to triggers reload


stefanpenner:
noticed it here: https://ember-twiddle.com/6a0c17f23fd8ce8ca95c?numColumns=1&route=%2Flocal-only

piotrpalek:
Hey I've made a PR to fix this issue, any comments on this are welcomed.

Btw. another approach I could think of (if we don't want to depend on the CodeMirror event) is to add a `fileContent` property to the `file-editor-column` component, pass that property to CodeMirror (instead of `file.content`) and then set `fileContent` to `file.content` on the `valueUpdated` event. This would allow us to know if the change came from within CodeMirrir (`fileContent` would  be different from `file.content`) or was set on the component externally.

Thoughts?

stefanpenner:
thanks, that PR looks good but likely needs a test. It should allow ember-twiddle to nicely address this issue. Good catch!

piotrpalek:
I was thinking of adding a test but I was wondering what I should test there. Would checking if I'm passing 3 arguments suffice? I would want to stay away from testing that for eg. the "instance" argument has some sort of properties since that should be covered in CodeMirror itself.

stefanpenner:
@piotrpalek ya, jsut a quick test ensuring all 3 args are passed should suffice.

Gaurav0:
Fixed by #297

stefanpenner:
:+1:

PR #297, opened 2015-12-19
Fix navigation reload

piotrpalek:
This fixes issue #290, but it depends on [these](https://github.com/IvyApp/ivy-codemirror/compare/19f4d62e4e6afe9b19c31fe919a7809ce827c56a...piotrpalek:pass-params) changes to codemirror.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "ember-moment": "4.2.1",
     "ember-notify": "4.3.1",
     "ember-responsive": "1.2.3",
-    "ivy-codemirror": "~1.2.0",
+    "ivy-codemirror": "piotrpalek/ivy-codemirror#pass-params",
     "torii": "^0.6.0"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "ember-moment": "4.2.1",
     "ember-notify": "4.3.1",
     "ember-responsive": "1.2.3",
-    "ivy-codemirror": "piotrpalek/ivy-codemirror#pass-params",
+    "ivy-codemirror": "1.3.0",
     "torii": "^0.6.0"
   },
   "dependencies": {
--- a/tests/integration/components/file-editor-column-test.js
+++ b/tests/integration/components/file-editor-column-test.js
@@ -52,7 +52,7 @@ test('it calls contentChanged with true when changing the content via the code e
   const enterKeyEvent = { keyCode: 3 };
 
   this.set('externalAction', (isUserChange) => {
-    assert.ok(isUserChange, 'contentChanged action was called');
+    assert.ok(isUserChange, 'contentChanged was called with isUserChange = true');
   });
 
   this.set('file', { content: '' });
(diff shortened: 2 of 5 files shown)


piotrpalek:
I've (temporarily) changed the CodeMirror dependency to the changed version so tests will pass on Travis.

piotrpalek:
I changed `contentsChanged` on the `file-editor-column` component, it didn't make sense that `contentsChanged` wouldn't fire when the content in the column actually changed. Now the gist controller gets to decide what to do depending on a flag the new action fires.

tests/integration/components/file-editor-column-test.js
```diff
@@ -2,54 +2,81 @@ import { moduleForComponent, test } from 'ember-qunit';
...
 
   this.$('.glyphicon-chevron-right').click();
+});
+
+test('it calls contentChanged with true when changing the content via the code editor', function(assert) {
+  assert.expect(1);
+  const enterKeyEvent = { keyCode: 3 };
+
+  this.set('externalAction', (isUserChange) => {
+    assert.ok(isUserChange, 'contentChanged action was called');
```
Gaurav0: I think you need to change the test description to match the new behavior.

tests/integration/components/file-editor-column-test.js
```diff
@@ -2,54 +2,81 @@ import { moduleForComponent, test } from 'ember-qunit';
...
+
+  const codeMirrorInstance = this.$('.CodeMirror')[0].CodeMirror;
+  codeMirrorInstance.triggerOnKeyDown(enterKeyEvent);
+});
+
+test('it calls contentChanged with false when changing the content programatically', function(assert) {
+  assert.expect(1);
+
+  this.set('externalAction', (isUserChange) => {
+    assert.notOk(isUserChange, 'contentChanged action was called');
```
Gaurav0: Here also.

Gaurav0:
This looks really good! Just need to fix up minor nits.

package.json
```diff
@@ -57,7 +57,7 @@
     "ember-moment": "4.2.1",
     "ember-notify": "4.3.1",
     "ember-responsive": "1.2.3",
-    "ivy-codemirror": "~1.2.0",
+    "ivy-codemirror": "piotrpalek/ivy-codemirror#pass-params",
```
rwjblue: It's ok if we can't, but I'd much prefer to make sure the upstream changes land before merging.

Gaurav0: Yes, please at least open a PR.

piotrpalek: Sure will do, it was just so you can see that the tests actually pass :)

tests/integration/components/file-editor-column-test.js
piotrpalek: I'm on mobile now so I could miss something, but what do you mean? The description is correct(imo) on the main test method above, and this assertion would say that it expected a true value (if it would fail). Did you mean to add that it expects a true value in this assertion?

Gaurav0: Instead of "contentChanged action was called", "contentChanged was called with isUserChange=true"

piotrpalek: Ah sure, I thought the title would be enough but I'll add it

package.json
piotrpalek: I've made a [PR](https://github.com/IvyApp/ivy-codemirror/pull/12) to CodeMirror

piotrpalek:
Changes landed upstream so I updated `package.json` to reflect that, if everything is all right should I rebase to a single commit?

Gaurav0:
No need, and actually I'm not a fan of rewriting history.

Gaurav0:
:+1:

piotrpalek:
:+1:

Issue #178, opened 2015-08-27
Support Tests


Gaurav0:
Allow users to write and execute Ember unit, integration, and acceptance tests.

rwjblue:
Should already be possible, but I'll try to whip up an example when I'm back at a keyboard...

Gaurav0:
https://ember-twiddle.com/4ecd2437e883e0bfe684?route=%2Ftests

Gaurav0:
Fixed by #298

PR #298, opened 2015-12-21
Make it easier to do Testing

Gaurav0:
This is for issue #178. We will still need some way in the UI to enable / disable testing "mode". Right now, it will be an option in twiddle.json, which also persists it.

@rwjblue @stefanpenner @joostdevries @alexspeller Please let me know what you think of my approach.

Screenshot:

screen shot 2015-12-21 at 3 04 57 pm


Gaurav0:
This PR is now complete. Please review.

PR #301, opened 2015-12-26
Fix file deletion

### LINKED ISSUE #237: Deleting files does not persist to previously saved gist.
Deleting files does not persist to an already saved gist.

Steps:

1. Visit http://ember-twiddle.com/
2. Save (no changes needed)
3. Delete `application/controller.js` file
4. Save again.
5. Review gist.github.com (or reload the twiddle)

After step 5. the actual saved gist still contains deleted files.

piotrpalek:
This fixes #237. The `registerDeleteOnGist` observer isn't always called, that's the reason why deleting doesn't work properly. 
I replaced it by overriding the `deleteRecord` method on the model which calls the neccesary code.
I'm assuming the observer didn't work because `isDeleted` is a computed property which isn't consumed anywhere, I tried fixing it by calling `this.get('isDeleted')` in the init method but that didn't really fix the issue completely (it seemed to help a little, didn't test it too much though).

Anyway I think depending on the `isDeleted` flag + observer might not be the best way to accomplish this to begin with.

### DIFF
--- a/app/models/gist-file.js
+++ b/app/models/gist-file.js
@@ -3,7 +3,7 @@ import Ember from 'ember';
 import Path from 'npm:path';
 
 const { attr, belongsTo } = DS;
-const { computed, observer } = Ember;
+const { computed } = Ember;
 
 export default DS.Model.extend({
   fileType: attr('string'),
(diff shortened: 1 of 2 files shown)


Gaurav0:
:+1:

PR #310, opened 2016-01-24
upgrade ember-cli/ember/ember-data to 2.3

piotrpalek:
Hey, so I wanted to do a quick upgrade of ember twiddles dependencies to the newest versions. Turned out a bit more complicated than anticipated, but I think I'm on the right path here.

I've backported the ember data shims from the previous ember-cli-shims version and include it by default in the twiddle app. Doesn't seem to cause any problems when using ember-data 2.3.0 and up, but I'm not sure about this so input is appreciated. If this could lead to potential problems I would add the shims conditionally when building the twiddle based on the ember-data version specified in twiddle.json.

This is currently still a WIP since I'm planning to remove container related deprecations as well, but as mentioned earlier wanted someone to take a look and check if my changes till now are fine.


Gaurav0:
Hi, @piotrpalek, yes your changes look basically fine.

Two questions:
1. Why was it necessary to abstract the resolver?
2. Can you conditionally include shims in the build only if ember data is less than 2.3?

piotrpalek:
Hey @Gaurav0, so:
1. I just followed the upgrade path for [ember-cli 2.3.0-beta.1](https://github.com/ember-cli/ember-cli/releases/tag/v2.3.0-beta.1). Or do you mean something else?
2. Yup as said in the PR description it's possible to provide them as a module. So I could import them in the ember-cli service, check for ED version and include them if necessary. Though as said they don't seem to cause any problems.

piotrpalek:
I still have 2 issues I would like to resolve:

1. Two tests are failing in Chrome (for eg. [this one](https://github.com/ember-cli/ember-twiddle/blob/master/tests/integration/components/file-editor-column-test.js#L50)) because the tests fire an additional [action](https://github.com/ember-cli/ember-twiddle/blob/master/app/components/file-editor-column.js#L37), which isn't fired when running in phantomjs. Any suggestions for this?
2. I've removed our own container related deprecation, but there are still a few left from addons. I'd like to check if there are newer versions that fix this issue and upgrade them as well if there are.

Gaurav0:
@piotrpalek Ah, didn't realize you had upgraded ember-cli as well.

I think the remaining issues can be done in a separate PR.

:+1:

piotrpalek:
@Gaurav0 fine with me to do that in another PR :smile:

PR #321, opened 2016-02-12
Update ember-cli to 2.3beta2

Gaurav0:
... also updating some other dependencies.


bower.json
```diff
@@ -4,24 +4,22 @@
     "ember": "2.3.1",
     "ember-cli-shims": "0.1.0",
     "ember-cli-test-loader": "0.2.2",
-    "ember-load-initializers": "0.1.7",
     "ember-qunit-notifications": "0.1.0",
-    "jquery": "^1.11.3",
+    "jquery": "1.11.3",
     "loader.js": "^3.5.0",
```
rwjblue: This should be removed I think...

vendor/loader-shim.js
```diff
@@ -0,0 +1,10 @@
+/* globals define */
+define('ember/load-initializers', ['exports', 'ember-load-initializers', 'ember'], function(exports, loadInitializers, Ember) {
```
rwjblue: Why do we need this shim?

bower.json
Gaurav0: Yup.

vendor/loader-shim.js
Gaurav0: I suppose i included it to be thorough in including all of ember-load-initializers in the (inner) app. But I guess it is not necessary.

Gaurav0:
@rwjblue Updated according to your comments.

rwjblue:
Awesome, thank you!

PR #324, opened 2016-02-16
Initial unit test blueprints

Gaurav0:
This makes it so easy to add a unit test for a controller, route, or service!


alexspeller:
:+1: looks good to me

Issue #328, opened 2016-02-17
Can we make it look more like an editor than JSBin?


taras:
I'd love to be able to see all of the files as soon as the gist opens. At the moment, we have 2 split screen, instead I'd prefer to see the files and 1 open file. 

<img width="1274" alt="screen shot 2016-02-17 at 4 31 54 pm" src="https://cloud.githubusercontent.com/assets/74687/13125543/17dfc9f2-d594-11e5-81df-0897beef57d3.png">

Can I haz dis?

alexspeller:
I actually think this is a better way to start with the files open, including all parts of the tree expanded - it's much easier to get a quick overview of what's in the twiddle that way. Slightly related, I'd personally like to be able to have a grid layout like jsfiddle instead of a column layout, as numerous small files are very common so it's be nice to see more at once.

taras:
We could probably break this up into several changes. Start with this and later add ability to open files in split screen mode. Something like: "Double click to split screen, single click to replace currently open"

Gaurav0:
Actually I was thinking if I should just make https://ember-twiddle.com?numFiles=1&fileTreeShown=true give @taras the effect he wanted.

Gaurav0:
@taras Is bookmarking the link above enough; can I close this issue?

taras:
I just checked. I'd prefer for that view to be default. Do you want me to get a vote going to see how many people would prefer to see the file list by defaut?

Gaurav0:
@taras Feel free.

taras:
Thank you @Gaurav0  :)

PR #330, opened 2016-02-18
Use query parameter to reflect whether file tree is shown

Gaurav0:
For issue #328


rwjblue:
:+1: but needs a rebase

PR #334, opened 2016-02-22
Change defaults to make twiddle look like an editor

### LINKED ISSUE #328: Can we make it look more like an editor than JSBin?

Gaurav0:
Fixes #328


alexspeller:
:+1:

Issue #66, opened 2015-07-22
Ability to open previous version of a gist


Gaurav0:

Gaurav0:
Fixed by #340

PR #340, opened 2016-03-01
Ability to load a previous revision of a gist

Gaurav0:
Issue #66


tests/acceptance/git-revision-test.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const files = [
+    {
+      filename: "application.template.hbs",
+      content: "Hello, World!"
+    }
+  ];
+
+  runGist({
+    type: "revision",
```
joostdevries: To me it looks weird to have a function called `runGist` which you can then pass an optional flag to run something else. Why not create `runGistRevision`?

app/serializers/gist-revision.js
```diff
@@ -0,0 +1,7 @@
+import GistSerializer from "./gist";
+
+export default GistSerializer.extend({
+  normalizeQueryResponse: function(store, primaryModelClass, payload, id, requestType) {
```
joostdevries: Out of curiousity, why is this needed?

Gaurav0: I'm querying for a record using two parameters. Ember Data expects an array back when you query. So I wrap the payload in an array.

tests/acceptance/git-revision-test.js
Gaurav0: You're right, it is weird. I did it because they were the mostly the same. I'll adjust. Thanks!

PR #343, opened 2016-03-04
Expose application instance to outside world

kkenan:


Gaurav0:
Hi @kkenan 

Did you test this? I think this should not work with a sandboxed iframe.

kkenan:
Hi. I've tested it with Chrome 48, Firefox 44 and Safari 8. I see no reason why it would not work. This content will be executed inside iframe's context so window object here is actually an iframe's contentWindow.

Gaurav0:
Tested your branch:

window.frames[0].contentWindow.App
Error: Permission denied to access property "contentWindow"

As I said, this should not work with a sandboxed iframe.

kkenan:
Oh. This PR was not supposed to expose application instance to outside of iframe. It just needed to be available inside the twiddle itself. I used initialisers to get app instance before, but I thought it's too burdensome. 

Never mind. Sorry for not being clear enough. :) Cheers. :+1:

Gaurav0:
@kkenan `getOwner` will give you the application instance.

PR #346, opened 2016-03-07
Add error message for an empty file

### LINKED ISSUE #338: Specific error message for fail to save gist when fresh file is empty.
I'm checking back in with twiddle and building out a mini site as a base to work with. I've been trying to build a basic 3 route page, for the last hour or two. I've written it out a few times but at the end, when I try to save, I get this error.

"Something went wrong. The gist was not saved."

I built it over again step by step and the same thing happened. _I'm stupid_, so I did it a 3rd time. The only thing I could think of, is that when I added my third route, I also added a template for that route – and there was no code in there. I added a single character, and it then saved properly. I'm sure it's more on the Gist end of things, and that it doesn't let you save _nothing_. 

Maybe templates are the only file that doesn't come with some code in place.

I could see this scaring people off for sure. A unique message would be helpful. "Could not save empty file to gist." or something along those lines. Another fix would be to just spit out an {{outlet}} in the template blue-print I guess.

Gaurav0:
Fixes #338

### DIFF
--- a/app/gist/route.js
+++ b/app/gist/route.js
@@ -104,6 +104,10 @@ export default Ember.Route.extend({
         this.get('notify').error("The gist is invalid, and could not be saved.");
         return;
       }
+      if (firstError.code === "missing_field") {
+        this.get('notify').error("The contents of a file is completely empty, so the gist could not be saved.");
+        return;
+      }
     }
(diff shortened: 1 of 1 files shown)


sheriffderek:
Excelent.

PR #347, opened 2016-03-08
Get more than 30 twiddles

Gaurav0:
Looks like gist.github.com is now only returning 30 gists by default. This increases that to a max of 100 gists. To do more, we will have to implement pagination.


stefanpenner:
:+1:

PR #349, opened 2016-03-09
[WIP] Addon support

joostdevries:
**VERY WIP**

Adds support for using addons in `twiddle.json` like this:
```json
{
  "addons": {
    "ember-tooltips": "latest"
  }
}
```

This works by requesting a combo of addon + addon version + ember version from an API which in turn redirects to an S3 bucket where the built addons are stored. If the addon is not built yet, it will trigger a build in the background (using lambda).

Build process:
- https://github.com/joostdevries/twiddle-backend/blob/master/addon-builder/ember-blueprints/1.13.15/ember-cli-build.js
- https://github.com/joostdevries/twiddle-backend/blob/master/addon-builder/index.js

Features:
- Will use NPM version resolving (so you can use "latests" or semver logic in `twiddle.json`)
- Will try to build the addon if it's not in our S3 bucket yet
- Will include both CSS + JS
- Will include app, addon and vendor JS + CSS

Todo
- [ ] A nice message while we're building
- [ ] A nice error message if the build failed
- [ ] Support multiple ember versions
- [ ] Backend fixes
  - [x] The current broccoli logic is a huge pile of hackery. Need to clean that up.
  - [x] Currently, we include more into the build then desired, for example we somehow include test assets in the liquid-fire build.
  - [ ] Support bower deps

### DIFF
--- a/app/services/dependency-resolver.js
+++ b/app/services/dependency-resolver.js
@@ -59,7 +59,7 @@ export default Ember.Service.extend({
 
   resolveAddon(name, value) {
     return new Ember.RSVP.Promise(function(resolve, reject) {
-      let url = `https://nl1fctyzr7.execute-api.us-east-1.amazonaws.com/staging/addon?addon=${name}&addon_version=${value}&ember_version=1.13.1`;
+      let url = `https://nl1fctyzr7.execute-api.us-east-1.amazonaws.com/staging/addon?addon=${name}&addon_version=${value}&ember_version=1.13.15`;
       $.getJSON(url, function(data){
         console.log(data);
         resolve(data);
--- a/app/services/dependency-resolver.js
+++ b/app/services/dependency-resolver.js
@@ -55,6 +55,7 @@ export default Ember.Service.extend({
         if(addon.status === 'build_success') {
           dependencies[name] = addon.addon_js;
           dependencies[name+'_css'] = addon.addon_css;
+          console.log(`Addon ${name} is loaded...`);
         }
         else if (addon.status === 'building') {
           console.log(`Addon ${name} is currently building...`);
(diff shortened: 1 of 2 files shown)


app/services/dependency-resolver.js
```diff
@@ -41,6 +41,38 @@ export default Ember.Service.extend({
     });
   },
 
+  resolveAddons: function(addons, dependencies) {
+    let addonPromises = {};
+    Object.keys(addons).forEach((name) => {
```
Gaurav0: This breaks if there is no addons hash in twiddle.json. You should return addonPromises if addons is null or undefined.

app/services/dependency-resolver.js
```diff
@@ -41,6 +41,38 @@ export default Ember.Service.extend({
...
+        }
+        else if (addon.status === 'build_error') {
+        }
+      });
+    });
+  },
+
+  resolveAddon(name, value) {
+    return new Ember.RSVP.Promise(function(resolve, reject) {
+      let url = `https://nl1fctyzr7.execute-api.us-east-1.amazonaws.com/staging/addon?addon=${name}&addon_version=${value}&ember_version=1.13.15`;
```
Gaurav0: Move url host and namespace to config/environment

Gaurav0:
Before tackling other things, you can

1. Merge with master
2. Resolve my comments
3. Get tests passing
4. Write an acceptance test using runGist

Thanks,
Gaurav

app/services/ember-cli.js
```diff
@@ -337,24 +336,28 @@ export default Ember.Service.extend({
...
-
-    var dependencyResolver = this.get('dependencyResolver');
-    dependencyResolver.resolveDependencies(twiddleJson.dependencies);
+      });
 
-    return twiddleJson;
+      var dependencyResolver = this.get('dependencyResolver');
+      dependencyResolver.resolveDependencies(twiddleJson.dependencies);
+      dependencyResolver.resolveAddons(twiddleJson.addons, twiddleJson.dependencies).then(() => {
+        console.log(twiddleJson);
```
Gaurav0: Please remove.

PR #374, opened 2016-03-28
Make dependencies order chronologically.

rwjblue:
Previously, the dependencies were ordered like:

screen shot 2016-03-28 at 13 59 57

After this change, the following order is used:

screen shot 2016-03-28 at 14 00 53


Gaurav0:
OK with me. :+1:

PR #388, opened 2016-04-07
Upgrade to Ember CLI 2.4.3

Gaurav0:


app/services/dependency-resolver.js
```diff
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 
-const EMBER_VERSIONS = ['2.4.3', '2.3.2', '2.2.2', '2.1.2', '2.0.3', '1.13.13', '1.12.2'];
+const EMBER_VERSIONS = ['2.4.4', '2.3.2', '2.2.2', '2.1.2', '2.0.3', '1.13.13', '1.12.2'];
```
rwjblue: Unrelated to this PR, it might be nice to not have to hard code this list.  Maybe we can come up with a way detect the list of valid versions (i.e. a `.json` file we can grab from builds.emberjs.com or something)...

blueprints/twiddle.json
```diff
@@ -9,8 +9,8 @@
   },
   "dependencies": {
     "jquery": "https://cdnjs.cloudflare.com/ajax/libs/jquery/1.11.3/jquery.js",
-    "ember": "https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.4.3/ember.debug.js",
+    "ember": "https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.4.4/ember.debug.js",
```
rwjblue: Can we use `"ember": "2.4.3"` (which is what you get if you select the version in the `dependencies` drop down) instead of the full URL?

blueprints/twiddle.json
```diff
@@ -9,8 +9,8 @@
...
     "ember-data": "https://cdnjs.cloudflare.com/ajax/libs/ember-data.js/2.4.3/ember-data.js",
-    "ember-template-compiler": "https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.4.3/ember-template-compiler.js"
+    "ember-template-compiler": "https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.4.4/ember-template-compiler.js"
```
rwjblue: Using `2.4.4` here might be nicer than the full URL...

blueprints/twiddle.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.7.1",
+  "version": "0.7.2",
```
rwjblue: As a separate issue, maybe we can have this automatically use the current ember-twiddle version?  I think this should be possible by using `git-repo-info` (or another related package) and add the `version` to the blueprint at runtime...

blueprints/twiddle.json
Gaurav0: Will do this in a separate PR. Want to test that properly before doing. Thanks for reviewing!

app/services/dependency-resolver.js
Gaurav0: Yes! I'll open an issue.

blueprints/twiddle.json
Gaurav0: No biggie. ember-cli-release auto updates this.

rwjblue: Ahh, nice!

PR #390, opened 2016-04-11
Do not use private arguments oldAttrs and newAttrs

Gaurav0:


app/components/main-gist.js
```diff
@@ -75,14 +75,16 @@ export default Ember.Component.extend(AppBuilderMixin, ColumnsMixin, FilesMixin,
   /**
    * reinitialize component when the model has changed
    */
-  didReceiveAttrs({ oldAttrs, newAttrs }) {
+  didReceiveAttrs() {
     this._super.apply(this, arguments);
 
-    if (!oldAttrs || (oldAttrs && newAttrs && oldAttrs.model.value !== newAttrs.model.value)) {
+    if (this.get('model') !== this._oldModel) {
```
rwjblue: Minor tweak, lets avoid `this.get('model')` twice:

```js
let model = this.get('model');

if (model !== this._oldModel) {

}

this._oldModel = model;
```

Gaurav0:
@rwjblue Fixed. Thanks for reviewing!

rwjblue:
Looks awesome!  IMO, this is much clearer (I may be biased though :smiling_imp:)...

PR #392, opened 2016-04-12
Update addon url

Gaurav0:


app/services/dependency-resolver.js
```diff
@@ -73,7 +74,7 @@ export default Ember.Service.extend({
 
   resolveAddon(name, value) {
     return new RSVP.Promise(function(resolve) {
-      const url = `https://nl1fctyzr7.execute-api.us-east-1.amazonaws.com/staging/addon?addon=${name}&addon_version=${value}&ember_version=1.13.15`;
+      const url = `${config.addonUrl}?ember_version=1.13.15&addon=${name}&addon_version=${value}`;
```
rwjblue: Shouldn't the `ember_version` be based on the current Ember version?

joostdevries: yas. but I have yet to create images for every ember version :-)

PR #393, opened 2016-04-12
Update Ember/Ember Data versions.

rwjblue:
Also updates default `twiddle.json` values to use simpler/nicer defaults.


rwjblue:
Hmm, I ran tests locally before pushing (and they passed).  I'll have to review in more detail in the morning...

rwjblue:
I removed the change to default the `twiddle.json` so that we can update the Ember and Ember Data versions first.  This is passing now...

Gaurav0:
:+1:

Issue #41, opened 2015-07-12
Change app name `demo-app` to `twiddle` in a backwards-compatible way


joostdevries:
Detect twiddle version and use either `demo-app` or `twiddle`.

--- 

2016-04-12

The app name was changed but ideally we make this backwards compatible. This requires:
- [ ] The addon-builder to use variable app names
- [ ] The UI to use variable app names

Gaurav0:
It would be even better if the user could specify.

joostdevries:
true. Maybe parse a minimal version package.json

joostdevries:
this sounds like a relatively simple thing now we have twiddle.json

leomeira:
Hey Gaurav, I would like to work on this issue.

Gaurav0:
@leomeira Thanks!

joostdevries:
I'd be careful with this considering the fact that addons actually need to know the app name so making it variable is probably not a good idea (or we'd have to write a custom resolver...)

Gaurav0:
Addons that rely on knowing the app name except via modulePrefix are buggy.

joostdevries:
@Gaurav0 that may be but during the build process anything in the `app` directory of an addon is namespaced to the app name. If we want #349 to work we need a consistent app (module) name.

Gaurav0:
I see. In that case let's agree on `app` and keep it simple?

leomeira:
:+1:

joostdevries:
The app name was changed but ideally we make this backwards compatible. This required
- [ ] The addon-builder to use variable app names
- [ ] The UI to use variable app names

joostdevries:
We may also choose NOT to do this, based on:

https://github.com/ember-cli/ember-twiddle/pull/394#issuecomment-208912214

> It doesn't break imports that use relative urls. We never made 'demo-app' name public.

joostdevries:
closed by #396

PR #394, opened 2016-04-12
Change app name to twiddle

Gaurav0:


joostdevries:
lgtm

rwjblue:
Hmm.  Does this break the addon processing?  I suspect that it needs to build the addons with the same app name...

joostdevries:
@rwjblue that's the reason for the change ;-)

rwjblue:
@joostdevries - And the cache of prebuilt addons was cleared?

joostdevries:
@rwjblue Yes. Because the build process was completely refactored to use docker instead of lambda (allowing git deps and muuuch faster build times and lower costs).

alexspeller:
Wait, am I missing something or will this break all existing twiddled that
have imports in them for local modules?

joostdevries:
@alexspeller the name was recently changed from `demo-app` to `app` so that breakage was already present at that point AFAIK.

rwjblue:
We should detect the twiddle.json version, and use the appropriate name.

joostdevries:
@rwjblue agreed but prefer to track that in isolated issue since breakage was already introduced.

Gaurav0:
@alexspeller It doesn't break imports that use relative urls.

joostdevries:
@rwjblue https://github.com/ember-cli/ember-twiddle/issues/41

rwjblue:
> so that breakage was already present at that point AFAIK

The change from `demo-app` to `app` was over 10 days ago (released in https://github.com/ember-cli/ember-twiddle/releases/tag/v0.7.2 5 days ago).  Anyone bit by this, likely now has to deal with this twice.

This tool is a large community resource, and should follow the same guidelines on breakages as Ember itself.  The change to `app` should have been reverted as soon as it was clear that this broke existing twiddles and the solution I proposed (and you created #41 to track) should have been implemented **before** introducing yet another breaking change.

joostdevries:
@rwjblue agreed. Let's hold off on releasing untill we fix #41

joostdevries:
or revert to `demo-app`

joostdevries:
@Gaurav0 we could also just do a very simple `replace(/demo-app/g, "twiddle")` in the UI only I guess

PR #398, opened 2016-04-13
Auto refresh after addon build completes

Gaurav0:
using power of ember-concurrency

### DIFF
--- a/app/services/dependency-resolver.js
+++ b/app/services/dependency-resolver.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import config from '../config/environment';
+import { task, timeout } from 'ember-concurrency';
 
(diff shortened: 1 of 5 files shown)


app/services/dependency-resolver.js
```diff
@@ -43,34 +44,50 @@ export default Ember.Service.extend({
...
 
   resolveAddons: function(addons, dependencies) {
-    let addonPromises = {};
-    Object.keys(addons).forEach((name) => {
-      var value = addons[name];
-
-      addonPromises[name] = this.resolveAddon(name, value);
+    const taskInstance = this.get('resolveAddonsTask').perform(addons, dependencies);
+    return taskInstance.then(() => {
+      return RSVP.resolve(task.value);
```
rwjblue: `task` is an import (from `ember-concurrency`), and is a method AFAIK (not a thing with a `.value` property)

Gaurav0: You are correct, and it should be `taskInstance.value`. I was trying to figure out why it still works, and it turns out I never used this value in the `then` where I called `resolveAddons`.

PR #411, opened 2016-04-25
make work \w npm v3

stefanpenner:


package.json
```diff
@@ -68,6 +68,7 @@
   "dependencies": {
     "babel": "^5.8.38",
     "babel-plugin-htmlbars-inline-precompile": "0.0.5",
+    "ember-cli-legacy-blueprints": "^0.1.0",
```
Gaurav0: Does this need to be in dependencies?

stefanpenner: typically we prefer this yes, as ember-cli being dependent on ^ may change over time. But this app appears to require it

stefanpenner: that way the dependencies are truthful and explicit .

Gaurav0: Should only be required during build. Moved to devDependencies.

stefanpenner: 👍

Issue #415, opened 2016-04-26
Don't hide reload button


locks:
How does the team feel about not hiding the reload button when live reload is selected? Sometimes I want to trigger a reload without having modified the code, and I have to deselect live reload to do it, or pointlessly edit the code.

alexspeller:
👍 I hit this surprisingly often

Gaurav0:
What is the use case for this?

sheriffderek:
I do this as well, but it more of a sanity check. As cool as CodePen and JSfiddle sandboxes are, they sometimes glitch and I just want to reload to be sure. That being said, if I could be sure that it's not the app - I would really enjoy not second guessing what the problem may be.

PR #417, opened 2016-04-26
Show loading animation while building

Gaurav0:


sheriffderek:
It seems like this could tie into #415 - and there could possibly be a reload 'button' that also acts as a loading / working spinner.

PR #431, opened 2016-05-10
Use custom model blueprint

pangratz:
Since the model blueprint from ember-cli-legacy-blueprints contains
custom logic for the import statements in the blueprints' index.js and
ember-twiddle doesn't support this (yet), we create our own model
blueprint inside ember-twiddle.

---

The currently generated model file using the blueprint from
ember-cli-legacy-blueprints looks like this (note the lacking import
for `Model`):

```js

export default Model.extend({
});
```

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -142,7 +142,6 @@ function getEmberCLIBlueprints() {
         'component-hbs': 'component/files/__root__/__templatepath__/__templatename__.hbs',
         'component-js': 'component/files/__root__/__path__/__name__.js',
         'controller': 'controller/files/__root__/__path__/__name__.js',
-        'model': 'model/files/__root__/__path__/__name__.js',
         'route': 'route/files/__root__/__path__/__name__.js',
         'service': 'service/files/__root__/__path__/__name__.js',
         'template': 'template/files/__root__/__path__/__name__.hbs',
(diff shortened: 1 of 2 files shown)


tests/unit/services/ember-cli-test.js
```diff
@@ -65,6 +65,21 @@ test("buildProperties() works as expected with replacements", function (assert)
   assert.ok(props.content.indexOf('helper(myHelper)') !== -1, 'Replacements worked if multiple');
 });
 
+test("buildProperties('model') creates a valid file for a model", function(assert) {
```
Gaurav0: This test is too brittle, breaking with any change to blueprints/model.js

I suggest removing.

ember-cli-build.js
```diff
@@ -178,5 +178,10 @@ function getEmberCLIBlueprints() {
   fileMap['index.html'] = fs.readFileSync('blueprints/index.html').toString();
   fileMap['test-start-app'] = fs.readFileSync('blueprints/start-app.js').toString();
 
+  // use custom model blueprint since the one currently defined in
+  // ember-cli-legacy-blueprints contains custom logic for import statements in
+  // the blueprints' index.js, which is not (yet) considered in ember-twiddle
```
Gaurav0: This long comment is not necessary, and it is fairly obvious that we are using a custom model.js

Gaurav0: Please also remove line for `model` in `legacyBlueprintFiles` hash. (This is overwriting it.)

tests/unit/services/ember-cli-test.js
pangratz: So you're suggesting not adding a test at all? I wouldn't expect too many changes for `blueprints/model.js` since all the basic stuff which is required to start implementing a model is there.

Just wanted to check before I remove the test...

ember-cli-build.js
pangratz: Okidoki

tests/unit/services/ember-cli-test.js
Gaurav0: Yes. If you'd like, you can instead write an acceptance test checking that a model can be added, in gist-test.js

pangratz: OK, will remove it then 😜

pangratz:
@Gaurav0 all set

PR #456, opened 2016-06-23
Update addon api endpoints

Gaurav0:


joostdevries:
I think you have to append `/addon` to both

Gaurav0:
Updated. Getting error:

```
XMLHttpRequest cannot load https://canary-addons.ember-twiddle.com/ember-1.13.15/ember-concurrency/0.6.0/addon.json. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access.
```

joostdevries:
should be fixed.

PR #470, opened 2016-08-01
Fix integration tests broken by upgrade to Ember 2.7

### LINKED ISSUE #468: Component integration tests broken
When trying to create a new component integration tests I'm getting the following errors:

screen shot 2016-07-29 at 1 50 13 pm

The files are generated correctly but twiddle doesn't enable testing mode so I can't run the tests.

Tested on Google Chrome  51.0.2704.103 (64-bit) and Safari 9.1.2 (11601.7.7)

Gaurav0:
The idea here is to use the same version of ember-testing.js as ember.js, just like with ember-template-compiler.js

Fixes #468

### DIFF
--- a/app/services/dependency-resolver.js
+++ b/app/services/dependency-resolver.js
@@ -18,6 +18,11 @@ const CDN_MAP = {
     fileName: 'ember-template-compiler.js'
   },
 
+  'ember-testing': {
+    library: 'ember.js',
+    fileName: 'ember-testing.js'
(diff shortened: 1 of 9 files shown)


rwjblue:
This looks good to me, thanks for tackling this so quickly @Gaurav0!

FWIW, this should allow folks to use ember-twiddle.com to prototype tests for 2.9.0-alpha issues.

PR #481, opened 2016-09-14
CSS fixes and improvements

piotrpalek:
The main issue I was trying to fix is that the "Run Tests" and "Live Reload" checkboxes cover each other when "Live Reload" is deselected:
Before:
image
After:
image

Other problems this will fix (discovered while fixing the first issue):
- the header elements are consistently positioned (especially vertically) across all panes (esp. visible for eg. the on "Output" string):
Before:
image
After:
image

- [when the tree is hidden and there are 0 editor panes](https://ember-twiddle.com/?fileTreeShown=false&numColumns=0) the page looked broken:
Before:
image
After:
image

- I've changed how the file tree looks and behaves, it is no longer resizable and is much smaller which leaves more space for the important stuff (it still expands when the tree shifts to the right due to many items):
Before:
image
After:
image

Mobile view (only last screen has real changes so I provide a "before" screenshot there):
image
image

Fixed output pane:
Before:
image
After:
image

Edit: I think I'd add a little more spacing to the file tree (left/right) looks a bit too cramped together imo

### DIFF
--- a/app/components/twiddle-panes.js
+++ b/app/components/twiddle-panes.js
@@ -14,9 +14,9 @@ export default Ember.Component.extend({
   didRender() {
     this._super(...arguments);
     if (!this.get('media.isMobile')) {
-      this.$('.twiddle-pane').after('<div class="handle"></div>');
+      this.$('.twiddle-pane.code, .twiddle-pane.output').after('<div class="handle"></div>');
       this.$('.handle').last().remove();
-      this.$('.handle').drags({ pane: '.twiddle-pane' });
+      this.$('.handle').drags({ pane: '.twiddle-pane.code, .twiddle-pane.output' });
     }
   },
 
--- a/app/components/twiddle-panes.js
+++ b/app/components/twiddle-panes.js
@@ -12,10 +12,11 @@ export default Ember.Component.extend({
   },
 
   didRender() {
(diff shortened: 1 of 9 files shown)


Gaurav0:
1. Please don't make the file tree unresizable. Making it smaller by default would be fine.
2. Please post mobile screenshots after changes.

piotrpalek:
@Gaurav0 Regarding #1 what's the utility of being able to resize the file tree?

piotrpalek:
@Gaurav0 I've updated the description with mobile screenhots. The changes are also visible [here](http://twiddlefork.herokuapp.com/)

piotrpalek:
@Gaurav0 noticed one more issue, the filepane behaves a bit weirdly when trying to shrink it below the minimum width (instead of nothing happening, the panel next to the filepane starts to grow). Let me fix that before merging.

piotrpalek:
@Gaurav0 I've fixed the tree issue and some additional ones:

- added a scrollbar when the filepanes contents are too big
Before:
image
After:
image

- the top toolbar was broken in mobile view when the text was too long (for eg. when displaying the revision)
Before:
image
After:
image

Gaurav0:
Looks Awesome. Thanks!

PR #495, opened 2016-11-20
Added underlines to saved twiddle links

acorncom:
I was expecting to be able to click the SHA and not having any luck.  I had no idea the Twiddle column and Gist column names were actually links I could click.  This makes it a little easier to see where the links are on the page:

screen shot 2016-11-19 at 7 28 08 pm

### DIFF
--- a/app/styles/_saved-twiddles.scss
+++ b/app/styles/_saved-twiddles.scss
@@ -14,4 +14,8 @@ div.table-responsive {
   tbody tr:hover {
     background-color: $pale-orange;
   }
+  
+  a {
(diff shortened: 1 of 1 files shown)


sheriffderek:
I noticed (or didn't notice) this as well.

Gaurav0:
Can we change the color as well while we are at it?

acorncom:
@Gaurav0 Done.  Here's what it looks like now:

screen shot 2016-12-21 at 8 17 25 pm

The Twiddle on the bottom row is currently being hovered over (although my cursor doesn't show up)

PR #506, opened 2017-01-07
adding github enterprise support

adhamhf:
This PR is to add github enterprise support so that ember-twiddle can be run privately if required.

Basically, this PR just replaces hard coded strings with reading from environmental variables.   I new torii provider for github enterprise was added because the baseUrl and redirectUrl are different for github enterprise distributions.

I talked to @Gaurav0 about this months ago.  Takes me a while to get PRs up due to company policy.

### DIFF
--- a/app/routes/gist.js
+++ b/app/routes/gist.js
@@ -1,10 +1,12 @@
 import Ember from 'ember';
+import config from '../config/environment';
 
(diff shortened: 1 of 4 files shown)


adhamhf:
Will reopen when I have fixed the issues.

PR #526, opened 2017-02-14
Enabling default keymap, and setting as default

alexander-alvarez:
I would like to hit `ctrl-arrow` when editing ember twiddles and have it work. :smiley: 

From the docs: http://codemirror.net/doc/manual.html
> keyMap: string
Configures the key map to use. The default is "default", which is the only key map defined in codemirror.js itself. Extra key maps are found in the key map directory. See the section on key maps for more information.

https://github.com/codemirror/CodeMirror/blob/master/src/input/keymap.js

### DIFF
--- a/app/models/settings.js
+++ b/app/models/settings.js
@@ -10,7 +10,7 @@ export default ObjectProxy.extend({
   storageKey: 'ember_twiddle_settings',
 
   defaultSettings: {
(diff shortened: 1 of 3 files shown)


Gaurav0:
LGTM. Thanks very much.

PR #528, opened 2017-02-24
Dropdown menu should be higher than handle

k-fish:
If the dropdown menu is open over the pane resize handle, the menu will unexpectedly close when the mouse reaches the handle.

This fix bumps the z-index back down (a change was made that increased the `z-index` of `.handle`) below that of  `.dropdown-menu`.

#### Example Behaviour
example_twiddle


k-fish:
Thanks 🎉

PR #538, opened 2017-03-23
Upgrade to ember-cli-qunit@3

Gaurav0:


Gaurav0:
This is waiting for a fix for https://github.com/qunitjs/qunit/issues/1119 before committing.

trentmwillis:
@Gaurav0 this should be safe to land now.

PR #548, opened 2017-05-05
removed redundant `RSVP.resolve`s

bekzod:
when value returned inside `then` callback there is no need to wrap it with `RSVP.resolve`  it is always resolved unless wrapped in `RSVP.reject`


Gaurav0:
LGTM

PR #575, opened 2018-01-14
Download project as zip file

nummi:
download


Gaurav0:
Very cool feature. Would it be possible to add some testing?

nummi:
@Gaurav0, I added to the file menu integration test — is that what you had in mind?

PR #578, opened 2018-02-10
Fix #557: Adds extra menu item for renaming a file

### LINKED ISSUE #557: Not clear how to rename files
I was giving a workshop and the attendees make a typo while creating a file. They then weren't sure how to rename the file, and only after a beat did I remember realize that "Move file" would do the trick.

kennethlarsen:
I've deicded to solve this issue by adding an extra menu item instead of writing something like "Move / Rename {{filename}}" or "Move or rename {{filename}} since I think it creates a lot of clutter. This won't help users figure out how to rename. 

Instead rename and move are seperate items of the menu with the same functionality. In my oppinion this provides a better user experience and more clearity.

### DIFF
--- a/app/templates/components/file-menu.hbs
+++ b/app/templates/components/file-menu.hbs
@@ -49,7 +49,7 @@
     </li>
     {{#if activeEditorCol}}
       <li><a {{action 'renameFile' activeFile}} class="test-rename-action">Rename {{activeFile.filePath}}</a></li>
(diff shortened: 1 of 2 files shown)


Gaurav0:
Thanks. Would you mind adding a quick test to https://github.com/ember-cli/ember-twiddle/blob/master/tests/integration/components/file-menu-test.js ?

PR #589, opened 2018-03-31
Adding docker support

jrock2004:
Adding support for people to use `Docker` to run the server.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,8 +15,8 @@ You will need the following things properly installed on your computer.
 * [Ember CLI](http://www.ember-cli.com/)
 * [PhantomJS 2](http://phantomjs.org/)
 * [Gatekeeper](https://github.com/prose/gatekeeper)
-* [Docker](https://docs.docker.com/install/)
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -15,13 +15,13 @@ You will need the following things properly installed on your computer.
 * [Ember CLI](http://www.ember-cli.com/)
 * [PhantomJS 2](http://phantomjs.org/)
 * [Gatekeeper](https://github.com/prose/gatekeeper)
+* [Docker](https://docs.docker.com/install/)
+* [Docker Compose](https://docs.docker.com/compose/install/)
```
Gaurav0: These should not be in the general list of prerequisites, as it is not necessary to use docker.

jrock2004:
Want to add optional text in there?

CONTRIBUTING.md
jrock2004: Sure I can figure something out

jrock2004:
Not sure why the tests would be failing as I changed no code. All I changed was the contribute file

Issue #593, opened 2018-04-05
Upgrade Babel to v6


Gaurav0:
This is likely to be a challenging upgrade, due to the differences in the Babel transpiler in v.5 vs v.6. 

Babel 5 transpiles modules using code here: https://github.com/ember-cli/ember-cli/blob/06080981a582ddefadaf16fbcbd2d3b3d397dee6/lib/models/addon.js#L1738-L1754

while Babel 6 transpiles modules using code here: https://github.com/ember-cli/ember-cli/blob/37fcddc7b75cd71ee148bee98fec603ce1516195/lib/broccoli/babel-process-modules-only.js#L18-L25

**Expected steps:**

- [ ] 1. Upgrade ember-twiddle to use Babel 6 for its own transpilation
- [ ] 2. Send to client Babel 6 runtime from ember-cli-build.js
- [ ] 3. Change how modules are compiled on client in [ember-cli.js](https://github.com/ember-cli/ember-cli/blob/37fcddc7b75cd71ee148bee98fec603ce1516195/lib/broccoli/babel-process-modules-only.js#L18-L25)
- [ ] 4. Get most tests working - ensure compilation is correct.
- [ ] 5. Ensure addons still work even if compiled with Babel 5
- [ ] 6. Fix hbs plugin
- [ ] 7. Ensure tests still work on client (enable_testing: true).
- [ ] 8. Fix remaining tests.

acorncom:
@Gaurav0 does the pR above mean this issue can already be closed? 👏

PR #594, opened 2018-04-05
Babel upgrade

Gaurav0:
For issue #593 

- [x] 1. Upgrade ember-twiddle to use Babel 6 for its own transpilation
- [x] 2. Send to client Babel 6 runtime from ember-cli-build.js
- [x] 3. Change how modules are compiled on client in [ember-cli.js](https://github.com/ember-cli/ember-cli/blob/37fcddc7b75cd71ee148bee98fec603ce1516195/lib/broccoli/babel-process-modules-only.js#L18-L25)
- [x] 4. Get most tests working - ensure compilation is correct.
- [x] 5. Ensure addons still work even if compiled with Babel 5
- [x] 6. Fix hbs plugin
- [x] 7. Ensure tests still work on client (enable_testing: true).
- [x] 8. Fix remaining tests.


package.json
```diff
@@ -18,7 +18,7 @@
   "license": "MIT",
   "devDependencies": {
     "broccoli-asset-rev": "^2.4.5",
-    "broccoli-babel-transpiler": "~5.6.1",
+    "broccoli-babel-transpiler": "~5.6.2",
```
rwjblue: It is slightly odd to leave this at 5.x, but it seems totally fine to update in a follow-up...

app/services/ember-cli.js
```diff
@@ -395,6 +395,13 @@ export default Ember.Service.extend({
       </script>`;
     }
 
+    depScriptTags += `
+      <script type="text/javascript">
+        Object.keys(requirejs.entries).forEach(function (moduleName) {
+          require(moduleName).__esModule = true;
```
rwjblue: Hmm, I don't think you should need to do this...

app/services/ember-cli.js
```diff
@@ -503,10 +510,16 @@ export default Ember.Service.extend({
...
   return {
-    modules: 'amdStrict',
+    presets: ['es2017'],
     moduleIds: true,
     moduleId: moduleName,
-    plugins: [ hbsPlugin ]
+    plugins: [
+      ['transform-es2015-modules-amd', {
+        loose: true,
+        noInterop: false
```
rwjblue: If you set this to `true`, I do not think that you will need the `__esModule = true` flag above (because Babel will not check if it has to interop).

package.json
Gaurav0: Fixed now.

app/services/ember-cli.js
Gaurav0: Added comment explaining

app/services/ember-cli.js
Gaurav0: Needed for compat with addons compiled with babel@5 using modules: 'amdStrict'

rwjblue:
👏 👏 👏 👏

PR #618, opened 2018-06-07
Run yarn upgrade and fix issues

Gaurav0:


tests/test-helper.js
```diff
@@ -20,7 +20,7 @@ Ember.Test.registerHelper('outputPane', function(app) {
 
 Ember.Test.registerHelper('outputContents', function(app, selector) {
   let output = app.testHelpers.outputPane();
-  return output.find(selector).text().trim();
+  return output.document.querySelector(selector).textContent.trim();
```
knownasilya: Should `output.document.querySelector(selector)` be checked for truthyness before using `textContent`? I think jquery handled this automatically before. Not sure..

Gaurav0: If it's not truthy a test will fail. ;)

knownasilya: Yeah, but the error won't be pretty haha

Issue #143, opened 2015-08-09
Add support for commenting/uncommenting


HeroicEric:
- [x] Allows commeting/uncommenting single lines
- [x] Allows commenting/uncommenting multiple lines

Looks like there is an addon for codemirror that adds support for this https://codemirror.net/doc/manual.html

Gaurav0:
I'm confused. Doesn't JavaScript let you do this?

HeroicEric:
@Gaurav0 I'm talking about inside the file editor panes. For example, on jsbin you can use `cmd + /` to comment/uncomment

joostdevries:
This should be fairly doalbe by adding the `comment` addon to codemirror (probably just `import` ing it). http://codemirror.net/doc/manual.html#addons

gokatz:
@Gaurav0 and @joostdevries, I would love to give this a try :)

PR #622, opened 2018-06-15
Adds shortcut option to comment javascript code.

gokatz:
- Adds shortcut option to comment javascript code.
  - `cmd-/` in sublime mode (codemirror default)
  - `alt-;` in emac mode (codemirror default)
  - `cmd-/` in default mode ([added in app](https://github.com/ember-cli/ember-twiddle/compare/master...gokatz:code-comment?expand=1#diff-2de9eda1655d499e934e7465e87ca0bdR6))
- Displays selected keystore type

w.r.t. handlebars, maintainer [just added meta data](https://github.com/codemirror/CodeMirror/commit/fbc002f2ba234b91417e086ff1cae0407320c158) to support commenting in handlebar extensions (not released). We can pull once `ivy-codemirror` updates the deps. 

Haven't implemented any tests for the change. Stuck with dynamic rendering of gist in test. Any help will be great.

related to #143

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,7 +31,8 @@ module.exports = {
     visit: true,
     waitForLoadedIFrame: true,
     waitForUnloadedIFrame: true,
-    '$': true
+    '$': true,
+    CodeMirror: false
   },
   rules: {
     'ember/new-module-imports': 'off'
--- a/app/app.js
+++ b/app/app.js
@@ -3,6 +3,12 @@ import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
+// adds "toggleCommentIndented" command to codemirror for default keymap
+
+if (CodeMirror) {
+  CodeMirror.keyMap.default['Cmd-/'] = 'toggleCommentIndented';
+}
+
 const App = Application.extend({
   modulePrefix: config.modulePrefix,
   podModulePrefix: config.podModulePrefix,
--- a/app/app.js
+++ b/app/app.js
@@ -3,12 +3,6 @@ import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
-// adds "toggleCommentIndented" command to codemirror for default keymap
-
-if (CodeMirror) {
-  CodeMirror.keyMap.default['Cmd-/'] = 'toggleCommentIndented';
-}
-
 const App = Application.extend({
   modulePrefix: config.modulePrefix,
   podModulePrefix: config.podModulePrefix,
--- a/tests/unit/initializers/app-initializer-test.js
+++ b/tests/unit/initializers/app-initializer-test.js
@@ -1,0 +1,26 @@
+import Application from '@ember/application';
+import { run } from '@ember/runloop';
+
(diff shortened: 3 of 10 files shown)


Gaurav0:
Thanks for this!

"Stuck with dynamic rendering of gist in test" - I don't think we need this, although the "runGist" helper is what I've been using.

I think we just need to
1. put anything into a codemirror
2. Simulate Ctrl-A
3. Simulate Ctrl-/
4. Check that the codemirror text has changed

Hope this helps.

gokatz:
Thanks @Gaurav0 I’m doing exactly the same. I had a doubt that whether the runGist helped can able to rerender on its own on content change.

Gaurav0:
Yes, it doesn't. However, we don't need to rerender the gist to create a test for this functionality.

gokatz:
> Simulate Ctrl-A

Need to use the `triggerKeyEvent` API which is present in `ember-cli-qunit` 4.3.x. While upgrading the package, I got an issue due to the custom test tree construction in our app as mentioned [here](https://github.com/ember-cli/ember-cli-qunit/issues/208#issuecomment-398017968). Tried using the `ember-cli-qunit` instead of `ember-qunit` and the build got succeeded. Can we proceed?  

> Check that the codemirror text has changed

Done! I thought of checking the rendered result.

app/app.js
```diff
@@ -3,6 +3,12 @@ import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
+// adds "toggleCommentIndented" command to codemirror for default keymap
+
+if (CodeMirror) {
+  CodeMirror.keyMap.default['Cmd-/'] = 'toggleCommentIndented';
+}
```
Gaurav0: Can we move this to somewhere more appropriate, like an initalizer?

.eslintrc.js
```diff
@@ -31,7 +31,8 @@ module.exports = {
     visit: true,
     waitForLoadedIFrame: true,
     waitForUnloadedIFrame: true,
-    '$': true
+    '$': true,
+    CodeMirror: false
```
Gaurav0: Why is this false?

gokatz: To access [here](https://github.com/ember-cli/ember-twiddle/pull/622/files#diff-2de9eda1655d499e934e7465e87ca0bdR8)

app/app.js
gokatz: 👍

gokatz:
🤦🏻‍♂️ initialiser test failed

tests/unit/initializers/app-initializer-test.js
```diff
@@ -0,0 +1,26 @@
...
+
+module('Unit | Initializer | app initializer', {
+  beforeEach() {
+    run(() => {
+      this.application = Application.create();
+      this.application.deferReadiness();
+    });
+  },
+  afterEach() {
+    destroyApp(this.application);
```
Gaurav0: This helper is for old "moduleForAcceptance" tests. I think you can remove it.

Gaurav0: nvm. It's in my code. You just have to edit the line in `tests/helpers/destroy-app` from `window.server.shutdown()` to `window.server && window.server.shutdown();`

Issue #658, opened 2019-06-18
Support files with multiple file endings


amiller-gh:
## Problem
Example: https://ember-twiddle.com/#/57b20467d824add6364cc47b3727d2dc

When ember-twiddle persists files to a gist, it replaces all `/`s (not allowed in gist file names) with `.`. The client makes an assumption that the all `.`s except the final one are actually `/`s. 

This assumption fails when a file has two file endings. Ex: a CSS Blocks `styles/stylesheet.block.css` file. When saved to a gist and re-read by the client, the filepath is interpreted incorrectly as `styles/stylesheet/block.css`.

## Proposal
To maintain backwards compatibility, but allow for multiple `.`s inside of a filepath, we can do the following:

On Save: 
 1. Escape all `.`s in the filepath with `\.` (allowed by gist)
 2. Replace all `/`s with an un-escaped `.`
 3. Persist to gist

On Read
 1. Replace all but the *final* un-escaped `.` with `/`
 2. un-escape any `\.`s, replacing with `.`
 3. Render

As gists are read and re-saved, their filepaths will be converted to:
`some.file.path.js` => `some.file.path\.js`

Because we still don't replace the final `.` with `/`, we maintain back compat.

New paths with more than one period will now be encoded properly in gist:
`my.file/path/to/a/file.mult.ext` => `my\.file.path.to.a.file\.mult\.ext`

amiller-gh:
@Gaurav0, PR posted 🙂 https://github.com/ember-cli/ember-twiddle/pull/659

PR #659, opened 2019-06-18
feat: Support files with multiple periods in the filepath.

amiller-gh:
fixes: https://github.com/ember-cli/ember-twiddle/issues/658

### DIFF
--- a/tests/acceptance/gist-test.js
+++ b/tests/acceptance/gist-test.js
@@ -47,6 +47,7 @@ test('deleting a gist loaded in two columns', function(assert) {
     click(secondFile);
     click(firstFilePicker);
     click(fileMenu);
+
     click(deleteAction);
     andThen(function() {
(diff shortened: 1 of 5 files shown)


tests/acceptance/gist-test.js
```diff
@@ -55,7 +55,7 @@ test('deleting a gist loaded in two columns', function(assert) {
     });
 
     // TODO: Replace brittle for loop test code with "while there are files left..."
-    for (var i = 0; i < 2; ++i) {
+    for (var i = 0; i < 1; ++i) {
```
amiller-gh: I'm not sure what was going on here, but it appears this didn't run on `head` either. It would delete the `twiddle.json` and cause a number of cascading test failures. Is `twiddle.json` supposed to be impossible to delete?

Gaurav0: yes, it is supposed to be impossible to delete, in that doing so recreates it.

PR #490, opened 2016-10-20
Redesign

knownasilya:
Implementation of #400 mockups by @matmac using ember-paper

Don't merge yet, this is a WIP. Goal is to get it as close to the mockups as possible, but not deviate too much from Ember Paper semantics.

#### TODO
- [x] Move to sidebar layout
- [x] Integrate ember-paper
- [x] Replace tree with https://github.com/IvyApp/ivy-tree
  * [ ] Fix expand all/collapse all buttons (expand all possible with '*' key, but otherwise see https://github.com/IvyApp/ivy-tree/issues/11)
- [x] Hide sidebar on desktop
- [x] Finish file menu
  * [x] Add file menu items
  * [x] Add rename/move commands (good for someone trying to help)
  * [x] Move gist operations to top (out of file menu) (good for someone trying to help)
  * [x] Figure out certain nesting issues
- [x] Update styles (colors, icons, fonts)
- [x] Mobile support ~(flexi?)~
  * Not using Flexi atm due to no Ember 2.10 support
  * [x] Finish sidebar layout
  * [x] Finish main screen layout when sidebar closed
- [ ] Fix tests
- [ ] Fix bugs

#### Another PR
- Move file operation onto tree
- Alternate layouts support
- Save user settings (keymap mode, layout mode, etc) to gist or localstorage
- Monaco editor integration

Screenshot updated 1/15/17

screen shot 2017-01-15 at 1 50 31 pm

PR #687, opened 2020-02-29
Sidebar

knownasilya:
Pulled out some of the initial changes from #490 to get it in.

twiddle-sidebar
Screen Shot 2020-02-29 at 5 11 10 PM
Screen Shot 2020-02-29 at 5 11 19 PM

### DIFF
--- a/app/components/file-tree.js
+++ b/app/components/file-tree.js
@@ -99,20 +99,16 @@ export default Component.extend({
       this.jsTreeActionReceiver.send('toggleNode', node.id);
     },
 
-    didBecomeReady() {
+    didBecomeReadyOptional() {
       if(this.didBecomeReady) {
         this.didBecomeReady();
       }
     },
 
-    didChange() {
+    didChangeOptional() {
       if (this.didChange) {
         this.didChange();
       }
-    },
-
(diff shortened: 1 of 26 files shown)


app/components/file-tree.js
```diff
@@ -99,22 +99,18 @@ export default Component.extend({
       this.jsTreeActionReceiver.send('toggleNode', node.id);
     },
 
-    didBecomeReady() {
+    didBecomeReadyOptional() {
```
Gaurav0: Ugh. Can't believe I wrote this. :)

app/styles/_toolbar.scss
```diff
@@ -14,70 +9,36 @@
     height: $topbar-height;
   }
 
-  @media (max-width: $screen-md-min) {
```
Gaurav0: Is this still responsive?

app/components/file-tree.js
knownasilya: We all have these moments :D

app/styles/_toolbar.scss
knownasilya: No, that is my next task haha.

PR #711, opened 2020-03-15
Remove redundent actions and use `{{on` and `(fn`

knownasilya:


Gaurav0:
@knownasilya status?

knownasilya:
Forgot about this, will try to get it across soon.

PR #735, opened 2020-10-06
fix broken switch test function

zabrinatan:
When toggling `Run Tests`, we are getting the below console error:
Screen Shot 2020-10-06 at 9 22 50 pm

On investigation, I found that `ensureTestStartAppHelperExists` and `ensureTestDestroyAppHelperExists` functions have been removed in this PR: https://github.com/ember-cli/ember-twiddle/pull/702/files#diff-ce45d0a320c9223a3cd12a5591a6584cL8-L19

This PR is to just remove the use of those two functions in the `switchTest` function as it's currently breaking this feature at the moment.


Gaurav0 (approved):
Thank you!

knownasilya:
Thanks!

PR #752, opened 2021-08-31
Stop checking for `-` in component file names

chriskrycho:
This restriction was lifted a long time ago, and is particularly out of step with Glimmer component usage.


knownasilya:
Looks like some tests need removing as well.

chriskrycho:
Yep. Figured that'd be the case – did the first change from the GitHub UI. 😂 Will finish this off later this week.

